PR #55, opened 2015-05-28
implement TypeToVar analyzer

### LINKED ISSUE #5: Implement TypeToVar analyzer
When a local variable is instantiated without `var`, offer a Code Fix to use `var` instead.

ragnarok56:
resolves #5


Vannevelj:
That looks really great! I also appreciate conforming to my unit test conventions and the solid amount of scenarios tested, thank you for that.

ragnarok56:
thanks, i try to match the style of the code that's already there, its easier to integrate then.  

I'm still used to not having new features, so I forget they are available.  i should fool around with that operator some more.

Something I had to manually update, and then revert before committing, was the post-build event in the main project, since it kept failing.  Maybe there's a better way to handle that?

Vannevelj:
Yes, there definitely should be a better way to do that. I will look into that issue and see if there's a better way for the NuGet post-build event.

Vannevelj:
@ragnarok56 see #56 for a fix. The default OutputPath is now `.` (aka: solution root) so you shouldn't have any errors anymore. You can also set your own path in the `NUGETLOCAL` environment variable you'll have to create yourself. I personally use this to keep all my local NuGet packages in one folder so I can group them in a single NuGet feed (I use this local feed to test everything works prior to publishing a new NuGet version).

Issue #77, opened 2015-06-01
Update tests to use RoslynTester v1.1.0-rc2


Vannevelj:
I released a new version of RoslynTester which has a more easy-to-use exposed API. There's also been some more cleaning up under the hoods and the error message should be improved in certain scenarios.

PR #78, opened 2015-06-01
Updated tests to use RoslynTester v1.1.0-rc2

Vannevelj:
Implements #77


ckuhn203:
That should make implementing tests a bit easier. :smile: :+1:

PR #83, opened 2015-06-03
Private Set AutoProperty can be ReadOnly AutoProperty Diagnostic

ckuhn203:
I squashed most of the commits to keep things clean, but there was a merge in there to fix the Nuget issue. I didn't have time to implement a quick fix for it, but it should certainly be possible, so I left some `//Todo:`s in there for it. I'm not sure I caught all the edge cases. So, if you can think of any other tests, I'd be happy to add them.


VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/PrivateSetAutoPropertyCanBeReadOnlyAutoProperty/PrivateSetAutoPropertyCanBeReadOnlyAutoPropertyAnalyzer.cs
```diff
@@ -0,0 +1,78 @@
...
+            {
+                context.ReportDiagnostic(diagnostic);
+            }
+        }
+
+        private Diagnostic HandleProperty(PropertyDeclarationSyntax propertyDeclaration)
+        {
+    
+            var setter = propertyDeclaration.AccessorList.Accessors.FirstOrDefault(s => s.IsKind(SyntaxKind.SetAccessorDeclaration) 
+                                                                                    && s.Modifiers.Any(m => m.IsKind(SyntaxKind.PrivateKeyword)));
```
Vannevelj: I believe you can use the `Modifiers.Any(SyntaxKind.PrivateKeyword)` overload here

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/PrivateSetAutoPropertyCanBeReadOnlyAutoProperty/PrivateSetAutoPropertyCanBeReadOnlyAutoPropertyAnalyzer.cs
```diff
@@ -0,0 +1,78 @@
...
+            if (setter != null)
+            {
+                if ( !TreeHasMethodsThatSetProperty(propertyDeclaration) )
+                {
+                    return Diagnostic.Create(Rule, setter.GetLocation(), "Property Setter", propertyDeclaration.Identifier);
```
Vannevelj: You're passing `Property Setter` and the identifier along but I don't see them used in your message. Am I overlooking something? I would indeed make the message more descriptive by including the name of the property it's about (You'll need the `.Text` or `.TextValue` property from the `Identifier` for that).

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/PrivateSetAutoPropertyCanBeReadOnlyAutoProperty/PrivateSetAutoPropertyCanBeReadOnlyAutoPropertyAnalyzer.cs
```diff
@@ -0,0 +1,78 @@
...
+            if (diagnostic != null)
+            {
+                context.ReportDiagnostic(diagnostic);
+            }
+        }
+
+        private Diagnostic HandleProperty(PropertyDeclarationSyntax propertyDeclaration)
+        {
+    
+            var setter = propertyDeclaration.AccessorList.Accessors.FirstOrDefault(s => s.IsKind(SyntaxKind.SetAccessorDeclaration)
```
Vannevelj: I'm not sure but can't you use `Accessors.OfType<SetAccessorDeclaration>().FirstOrDefault()` here?

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/PrivateSetAutoPropertyCanBeReadOnlyAutoProperty/PrivateSetAutoPropertyCanBeReadOnlyAutoPropertyAnalyzer.cs
```diff
@@ -0,0 +1,78 @@
...
+                    return Diagnostic.Create(Rule, setter.GetLocation(), "Property Setter", propertyDeclaration.Identifier);
+                }
+            }
+
+            return null;
+        }
+
+        private static bool TreeHasMethodsThatSetProperty(PropertyDeclarationSyntax propertyDeclaration)
+        {
+            var root = propertyDeclaration.SyntaxTree.GetRoot();
```
Vannevelj: Do we need to look through the entire `root`? Since it's about `private set;`, I think we can just look through the outer-most class (to account for nested classes). For example this:

	class X
	{
		public static int A { get; private set; }
		
		class Y
		{
			Y()
			{
				A = 5;
			}
		}
	}

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/PrivateSetAutoPropertyCanBeReadOnlyAutoProperty/PrivateSetAutoPropertyCanBeReadOnlyAutoPropertyAnalyzer.cs
```diff
@@ -0,0 +1,78 @@
...
+            var assignments = root.DescendantNodes().OfType<AssignmentExpressionSyntax>();
+                                
+            return assignments.Select(a => a.Left)
+                                .Cast<MemberAccessExpressionSyntax>()
+                                .Select(l => l.Name.Identifier)
+                                .Any
+                                (
+                                    i => i.Text == propertyDeclaration.Identifier.Text
+                                    && i.Parent.Ancestors().All(anc => anc.GetType() != typeof(ConstructorDeclarationSyntax))
```
Vannevelj: Is it possible that `i.Parent` itself already is of type `ConstructorDeclarationSyntax`? If so, using `.AncestorsAndSelf()` is probably more appropriate otherwise that scenario would slip through the cracks.

ckuhn203: No. Not possible. A `MemberAccessExpression` couldn't possibly be a `ConstructorDeclarationSyntax`.

Vannevelj: But `MemberAccessExpressionSyntax` refers to `i`. Here you use `i.Parent.Ancestors()`. There is no guarantee here what `i.Parent` is, right?

ckuhn203: Right. I guess what I'm not understanding why `.AncestorsAndSelf()` rather than just `.Ancestors()`.

Vannevelj: In case `i.Parent` is the `ConstructorDeclarationSyntax`. If you use `Ancestors()` it will look at `i.Parent`'s nodes above it. If you use `AncestorsAndSelf()` it will also include `i.Parent` (which may be the constructor already).

ckuhn203: Right. That's the intended behavior. If a Private Set Autoproperty is set in the constructor, it shouldn't be reported. Maybe I'm still missing something though. I'll ping you in chat when I get a chance.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/PrivateSetAutoPropertyCanBeReadOnlyAutoProperty/PrivateSetAutoPropertyCanBeReadOnlyAutoPropertyAnalyzer.cs
ckuhn203: Yup. I can. That looks much nicer.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/PrivateSetAutoPropertyCanBeReadOnlyAutoProperty/PrivateSetAutoPropertyCanBeReadOnlyAutoPropertyAnalyzer.cs
ckuhn203: I'm honestly unsure of what that line of code really does and what you're looking for. I'm very guilty of some copy pasta here.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/PrivateSetAutoPropertyCanBeReadOnlyAutoProperty/PrivateSetAutoPropertyCanBeReadOnlyAutoPropertyAnalyzer.cs
ckuhn203: As best as I can tell, there is only an `AccessorDeclartionSyntax` that covers both getter and setter. `SetAccessorDeclartion` is a member of the `SyntaxKind` enum, so this won't work.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/PrivateSetAutoPropertyCanBeReadOnlyAutoProperty/PrivateSetAutoPropertyCanBeReadOnlyAutoPropertyAnalyzer.cs
Vannevelj: For completeness' sake: you could also use `Accessors.OfType<SetAccessorDeclaration>().FirstOrDefault()`

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/PrivateSetAutoPropertyCanBeReadOnlyAutoProperty/PrivateSetAutoPropertyCanBeReadOnlyAutoPropertyAnalyzer.cs
Vannevelj: `Diagnostic.Create()` just creates a diagnostic (yes, really ;) )and takes 3 arguments: the first on is the `DiagnosticRule` we defined above which contains message, severity, ID, etc. The second argument is the location in the document where we want to show our warning (the squiggly lines you see in the IDE). The third argument is of type `params object[]` and contains the data you want to format your message with. If you'll look at other analyzers that use this third argument, you'll see that their message looks like this: `const string Message = "{0} blaat {1}`. It's the same idea as `string.Format()`: you replace the placeholders there with the data you pass in as arguments.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/PrivateSetAutoPropertyCanBeReadOnlyAutoProperty/PrivateSetAutoPropertyCanBeReadOnlyAutoPropertyAnalyzer.cs
Vannevelj: You're right, I'm mistaken here.

ckuhn203:
I pushed the `autoproperties` branch into this repo. There's an issue with picking up assignments without the `this` keyword. I noted it. I spent a good part of the morning trying to fix it, but it was like playing whack a mole. I'd get one failing test green, only to break another. 

Anyway, I'm going to kill this PR.

Issue #20, opened 2015-05-03
Implement SingleEmptyConstructor analyzer


Vannevelj:
If there is only one `public` constructor which is parameterless and does not contain code inside, offer to remove it. 

Discussion note: if there are comments inside the constructor's body, I would suggest to not display a warning.

Vannevelj:
Fixed in #113 by @Hosch250

PR #112, opened 2015-08-05
Close issue #20 - Find and remove public empty ctor's with no comments.

Hosch250:


Vannevelj:
Could you also make it account for attributes? Otherwise default constructors that have attributes shouldn't show a warning.

PR #137, opened 2015-08-09
Close #87

### LINKED ISSUE #87: Implement NoExplicitAccessModifier analyzer
Give each field, property, class, etc their explicit access modifier. Nobody wants to remember that.. Case in point: research what all the default access modifiers are and update this issue.

Hosch250:
Close #87


VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ExplicitAccessModifiers/ExplicitAccessModifiersAnalyzer.cs
```diff
@@ -0,0 +1,221 @@
...
+                    return "protected internal";
+                case Accessibility.Protected:
+                    return "protected";
+                case Accessibility.Internal:
+                    return "internal";
+                case Accessibility.Public:
+                    return "public";
+                default:
+                    // friend has the same value as internal, and not applicable isn't an option here
+                    return "something happened - please create an issue on our GitHub page";
```
Vannevelj: Throw an exception instead, this would generate a general error anyway since it wouldn't be able to parse the tree. Something like `throw new ArgumentException(nameof(accessibility));`

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ExplicitAccessModifiers/ExplicitAccessModifiersAnalyzer.cs
```diff
@@ -0,0 +1,221 @@
...
+
+        private readonly SyntaxKind[] _modifierKinds =
+        {
+            SyntaxKind.PublicKeyword,
+            SyntaxKind.ProtectedKeyword,
+            SyntaxKind.InternalKeyword,
+            SyntaxKind.PrivateKeyword
+        };
+
+        private string AccessibilityToString(Accessibility accessibility)
```
Vannevelj: I'm leaning towards simply using `accessibility.ToString().ToLower()`

PR #153, opened 2015-08-14
Fix several crash-causing bugs

### LINKED ISSUE #149: Explicit access modifier should be first in a method's modifier list
When applying the "ExplicitAccessModifier" code fix, it puts the access modifier behind other modifiers like `static`. The access modifier should be the very first in line.

For example we go from this:

        static void Main(string[] args)
        {
        }

to this:

        static private void Main(string[] args)
        {
        }

but it should be this:

        private static void Main(string[] args)
        {
        }

### LINKED ISSUE #151: ObsoleteAttributeWithoutReason crashes when you begin typing an attribute
Given a class like this

    class MyClass
    {
    }

as soon as you begin to type an attribute, it will crash because of ObsoleteAttributeWithoutReasonAnalyzer.cs:39.

The following line throws the exception because at this point there is a half-finished attribute, meaning the symbol is `null`.

    var type = context.SemanticModel.GetSymbolInfo(attributeExpression).Symbol;
    if (type.ContainingType.MetadataName != typeof (ObsoleteAttribute).Name)

### LINKED ISSUE #152: SingleEmptyConstructor crashes when you define a constructor
While typing the constructor, as soon as you close the `parameterList` it crashes on SingleEmptyConstructorAnalyzer.cs:51:

    if (constructorDeclaration.Body.Statements.Any())

This happens because `Body` is `null` at this point.

Hosch250:
Close #149 
Close #151 
Close #152

### DIFF
--- a/VSDiagnostics/VSDiagnostics/VSDiagnostics.Test/Tests/Attributes/EnumCanHaveFlagsAttributeTests.cs
+++ b/VSDiagnostics/VSDiagnostics/VSDiagnostics.Test/Tests/Attributes/EnumCanHaveFlagsAttributeTests.cs
@@ -1,0 +1,110 @@
+﻿using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using RoslynTester.Helpers.CSharp;
+using VSDiagnostics.Diagnostics.Attributes.EnumCanHaveFlagsAttribute;
+
+namespace VSDiagnostics.Test.Tests.Attributes
+{
+    [TestClass]
+    public class EnumCanHaveFlagsAttributeTests : CSharpCodeFixVerifier
+    {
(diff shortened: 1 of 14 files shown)


VSDiagnostics/VSDiagnostics/VSDiagnostics.Test/Tests/Attributes/EnumCanHaveFlagsAttributeTests.cs
```diff
@@ -0,0 +1,139 @@
...
+    enum Foo
+    {
+    }
+}";
+
+            VerifyDiagnostic(original, EnumCanHaveFlagsAttributeAnalyzer.Rule.MessageFormat.ToString());
+            VerifyFix(original, result);
+        }
+    }
+}
```
Vannevelj: Can you add a test that verifies it doesn't add the attribute when it's part of a chained attribute like this: `[Obsolete("reason"), Flags]`. I'm pretty sure it won't give a problem with the current implementation, but I'd like to have this scenario covered as well. Likewise a test that uses `[FlagsAttribute]` would be useful as well.

Issue #90, opened 2015-06-08
Implement StringPlaceholdersInWrongOrder analyzer


Vannevelj:
The numerical placeholders in a string should count upwards from 0 to n - 1.

Go from this:

    string.Format("abc {2} def {0} ghi {1}", x, y, z);

to this:

    string.Format("abc {0} def {1} ghi {2}", z, x, y);

Hosch250:
This should be tested with this use case also:

    string.Format("abc {0} def {1} ghi {2} def {1} abc {0}", y, z, x);

PR #158, opened 2015-08-15
Issue90

Vannevelj:
Implements #90


VSDiagnostics/VSDiagnostics/VSDiagnostics.Test/Tests/Strings/StringPlaceholdersInWrongOrderTests.cs
```diff
@@ -0,0 +1,457 @@
...
+    using System;
+    using System.Text;
+
+    namespace ConsoleApplication1
+    {
+        class MyClass
+        {
+            void Method()
+            {
+                string s = string.Format(""Hello {1}, my name is {2}. Yes you heard that right, {2}."", ""Mrs. Testing"", ""Mr. Tester"", ""Mr. Test"");
```
Vannevelj: This isn't correct. The placeholder starts at 1 but the used arguments start at 0.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Strings/StringPlaceholdersInWrongOrder/StringPlaceholdersInWrongOrderAnalyzer.cs
```diff
@@ -0,0 +1,116 @@
...
+
+            var firstArgumentSymbol = context.SemanticModel.GetSymbolInfo(firstArgument.Expression);
+            if (!(firstArgument.Expression is LiteralExpressionSyntax) &&
+                (firstArgumentSymbol.Symbol?.MetadataName == typeof(CultureInfo).Name && !(secondArgument?.Expression is LiteralExpressionSyntax)))
+            {
+                return;
+            }
+
+            // Get the formatted string from the correct position
+            var firstArgumentIsLiteral = firstArgument.Expression is LiteralExpressionSyntax;
```
SLaks: Instead of assuming a literal, call http://source.roslyn.io/#Microsoft.CodeAnalysis/Compilation/SemanticModel.cs,718

This will make it work with constants, parentheses, and concatenation.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Strings/StringPlaceholdersInWrongOrder/StringPlaceholdersInWrongOrderAnalyzer.cs
```diff
@@ -0,0 +1,116 @@
...
+            // The overloads are in the form string.Format(string, object[]) or string.Format(CultureInfo, string, object[])
+            if (invocation.ArgumentList == null || invocation.ArgumentList.Arguments.Count < 2)
+            {
+                return;
+            }
+
+            var firstArgument = invocation.ArgumentList.Arguments[0];
+            var secondArgument = invocation.ArgumentList.Arguments[1];
+
+            var firstArgumentSymbol = context.SemanticModel.GetSymbolInfo(firstArgument.Expression);
```
SLaks: Instead of checking the type of the first argument, just find whichever argument corresponds to the argument named `format`.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Strings/StringPlaceholdersInWrongOrder/StringPlaceholdersInWrongOrderAnalyzer.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        private void AnalyzeNode(SyntaxNodeAnalysisContext context)
+        {
+            var invocation = context.Node as InvocationExpressionSyntax;
+            if (invocation == null)
+            {
+                return;
+            }
+
+            // Verify we're dealing with a string.Format() call
```
SLaks: Instead of hard-coding for `string.Format`, look for any invocation of a method which takes a `string format` followed by either a `params object[]` or a series of `object`s.

This will also catch methods like `Console.WriteLine`, `StreamWrite.Write`, `Debug.WriteLine`, etc.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Strings/StringPlaceholdersInWrongOrder/StringPlaceholdersInWrongOrderAnalyzer.cs
Vannevelj: I've been looking into this but it's not entirely clear to me how this would be done. How do I find which argument corresponds to a particular parameter?

SLaks: By index.  To get the parameters, cast the symbol to `IMethodSymbol` and look at parameters.

Vannevelj: Ahhh it just clicked. When it looks up the method symbol, method resolution will already have happened so it will already have the correct method overload. Seems like this should do it just fine then:

```csharp
Debug.Assert(invocation.Expression is MemberAccessExpressionSyntax);
var memberAccessExpression = (MemberAccessExpressionSyntax) invocation.Expression;
var invokedMethod = context.SemanticModel.GetSymbolInfo(memberAccessExpression.Name);
var methodSymbol = invokedMethod.Symbol as IMethodSymbol;

var formatParam = methodSymbol?.Parameters.FirstOrDefault(x => x.Name == "format");
if (formatParam == null)
{
	return;
}

var formatIndex = formatParam.Ordinal;
```

SLaks: You should get the symbol from the `InvocationExpression` (which also has the same symbol), to work with `using static`.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Strings/StringPlaceholdersInWrongOrder/StringPlaceholdersInWrongOrderAnalyzer.cs
Vannevelj: I'm a little hesitant about this one since it increases the risk of false positives. What do you think about adding multiple checks like this:

    if (!invocation.IsAnInvocationOf(typeof(string), nameof(string.Format), context.SemanticModel))
     if (!invocation.IsAnInvocationOf(typeof(Console), nameof(Console.WriteLine), context.SemanticModel))

etc?

SLaks: I don't think it's very likely that you'd have a method that takes `string format` and multiple `object`s, but is not a format string.
`object` parameters should be rare, and `format` as a parameter name has few other meanings.

Vannevelj: I have implemented your suggestions in my PR for the feature that I will use to discuss on my paper: #319

SLaks: You can also talk about these pitfalls (especially language corner cases like named parameters, non-member-access invocations, and concatenated constants) in the paper.

Vannevelj: Definitely. Creating an analyzer is easy -- creating an analyzer that adheres to the entire language specification is incredibly hard

SLaks: Read through http://source.roslyn.io/#Microsoft.CodeAnalysis.CSharp.Workspaces/Extensions/ExpressionSyntaxExtensions.cs for a fascinating collection of methods that handle these kind of corner cases.

Vannevelj:
At first glance that looks really interesting -- I'll take that with me.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Strings/StringPlaceholdersInWrongOrder/StringPlaceholdersInWrongOrderAnalyzer.cs
Vannevelj: At first glance that looks really interesting -- I'll take that with me.

PR #201, opened 2015-08-19
Fixes #198

### LINKED ISSUE #198: NullReferenceException when using object initializer
Introducing code that using an object initializer, throws an exception on ArgumentExceptionWithoutNameOfAnalyzer:39.

The reason for this is because the `ArgumentList` property here is `null` (this would not be null if there would be braces after the `new T()` expression.

I believe this is a fairly common mistake that we made before, so look for other occurrences of this problem.

Vannevelj:


Vannevelj:
Couldn't replicate the issue with a test but I verified manually

PR #216, opened 2015-08-21
Close #215

### LINKED ISSUE #215: If without braces on `else if () {}`
This code fires the If Without Braces inspection because the `else` doesn't have braces.

    if (condition)
    {
        //
    }
    else if(condition)
    {
        //
    }

Hosch250:


Vannevelj:
Can you also add a case for 

	if(5 == 5)
	{
		Console.WriteLine();
	} else if(6 == 6)
		Console.WriteLine();

I have a feeling this might've been overlooked

PR #220, opened 2015-08-22
Close #217

### LINKED ISSUE #217: Single empty constructor with a base call still triggers
This displays a warning:

    public ExcelApp() : base("Excel") { }

A call to `base()` can trigger a warning but not when it takes arguments.

Hosch250:


Vannevelj:
can you add tests that show using `: this ()` works as well? (with and without args)

Issue #258, opened 2015-09-14
Implement DoNotMutateStructFromSelf


Vannevelj:
Structs shouldn't be able to mutate themselves. They shouldn't be able to be mutated at all, but this specific scenario deserves its own diagnostic.
Refer to the [struct guidelines](https://msdn.microsoft.com/en-us/library/ms229031(v=vs.110).aspx) for more details. 

An example of such violating code:

	void Main()
	{
		var obj = new X(10);
		obj.Mutate();
	}
	
	struct X
	{
		private int x;
		
		public X(int value)
		{
			x = value;
		}
	
		public void Mutate()
		{
			this = new X(5);
			Console.WriteLine(this.x);
		}
	}

Vannevelj:
Implemented by @nemec in #260

PR #260, opened 2015-09-20
Add diagnostic for #258, a struct mutating itself.

nemec:
I ended up adding a new diagnostic category for "structs", given that there are a few more open issues regarding structs, but I'm not sure what your policy is regarding category assignment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ Currently these diagnostics are implemented:
 | General | UseAliasesInsteadOfConcreteType | Use the built-in type aliases instead the concrete type.
 | Strings | ReplaceEmptyStringWithStringDotEmpty | Use `string.Empty` instead of `""`.
 | Strings | StringPlaceholdersInWrongOrder | Adjusts the placeholders in `string.Format()` calls to be in numerical order.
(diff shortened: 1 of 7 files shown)


Hosch250:
Looks nice.  Generally we would use the "General" category for this, but I've been thinking lately that we should add a few more categories.  If you wish to chat about this project, we have a chatroom setup at Code Review.SE: http://chat.stackexchange.com/rooms/26639/vsdiagnostics

Vannevelj:
Looks good! The added category is fine -- maybe we should look into creating more categories anyway.

Would you mind adding a few more tests to account for different execution paths? I'm thinking like for example assiging `new X(5)` to a variable instead of `this`. This will prevent us from creating problems in the future should that diagnostic ever need refactoring.

nemec:
@Hosch250 bookmarked! I'll hang around.

@Vannevelj you're thinking of additional negative tests? Sure, I'll add a few of those and get back to you.

nemec:
Added more tests to the commit.

Vannevelj:
Great additions! This looks good and will be included in the next release. I appreciate the effort you've undertaken, @nemec

Issue #23, opened 2015-05-03
Implement SyncMethodWithAsyncKeyword analyzer


Vannevelj:
Synchronous methods should not have "Async" as suffix.

Vannevelj:
Implemented in #301 by @tiesmaster

PR #301, opened 2015-11-16
Implementation of #23: SyncMethodWithAsyncSuffix

tiesmaster:
As discussed in the stackexchange chat, I think this analyzer could easily be merged with the [`AsyncMethodWithoutAsyncSuffixAnalyzer`](https://github.com/Vannevelj/VSDiagnostics/blob/develop/VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Async/AsyncMethodWithoutAsyncSuffix/AsyncMethodWithoutAsyncSuffixAnalyzer.cs), but as @Hosch250 was indicating, we can always do that later. Maybe, the logic for determining if the method is async is a good fit for an extension method.

If there is anything wrong with the formatting, please let me know. I couldn't find a style guide, or something like that.


VSDiagnostics/VSDiagnostics/VSDiagnostics.Test/Tests/Async/SyncMethodWithAsyncSuffixTests.cs
```diff
@@ -0,0 +1,637 @@
...
+{
+    [TestClass]
+    public class SyncMethodWithAsyncSuffixTests : CSharpCodeFixVerifier
+    {
+        protected override CodeFixProvider CodeFixProvider => new SyncMethodWithAsyncSuffixCodeFix();
+
+        protected override DiagnosticAnalyzer DiagnosticAnalyzer => new SyncMethodWithAsyncSuffixAnalyzer();
+
+        [TestMethod]
+        public void SyncMethodWithAsyncSuffix_WithoutAsyncKeywordAndNoSuffix_DoesNotDisplayWarning()
```
Vannevelj: We've changed our "convention" and no longer suffix our tests with `DoesNotDisplayWarning` or `DisplaysWarning` -- can you remove these? The intent of the test is always clear from the way we verify diagnostics and there are only two possible outcomes: a failure of a test always requires us to dive into it anyway.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Async/SyncMethodWithAsyncSuffix/SyncMethodWithAsyncSuffixAnalyzer.cs
```diff
@@ -0,0 +1,109 @@
...
+            {
+                if (method.Identifier.Text.EndsWith("Async"))
+                {
+                    context.ReportDiagnostic(Diagnostic.Create(Rule, method.Identifier.GetLocation(),
+                        method.Identifier.Text));
+                }
+            }
+        }
+
+        private static bool IsDefinedInAncestor(IMethodSymbol methodSymbol)
```
Vannevelj: This is the same method as in https://github.com/Vannevelj/VSDiagnostics/blob/ccf2ba76e6cb93873b297bee6970dd184ef7a5ff/VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Async/AsyncMethodWithoutAsyncSuffix/AsyncMethodWithoutAsyncSuffixAnalyzer.cs#L75 right? These should be moved to an extension method in util/Extensions.cs so we don't duplicate this behaviour. This is even more important because if I recall correctly, there is a scenario in which this method fails (but I haven't looked too deep into it yet -- there's a bug report filed).

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Async/SyncMethodWithAsyncSuffix/SyncMethodWithAsyncSuffixAnalyzer.cs
```diff
@@ -0,0 +1,109 @@
...
+                var interfaceMethods = @interface.GetMembers().Select(type.FindImplementationForInterfaceMember);
+                if (interfaceMethods.Any(method => method.Equals(methodSymbol)))
+                {
+                    return true;
+                }
+            }
+
+            // Start with the first ancestor
+            type = type.BaseType;
+            while (type != null)
```
Vannevelj: This is probably the line that causes an issue: `typeof(object).BaseType` will never be `null`: it's a circular reference. But don't worry about that, there's a separate issue for that already.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Utilities/DiagnosticId.cs
```diff
@@ -42,5 +42,6 @@ public static class DiagnosticId
         public const string TestMethodWithoutPublicModifier = "VSD0038";
         public const string FlagsEnumValuesDontFit = "VSD0039";
         public const string NamingConventionsConflictingMember = "VSD0040";
+        public const string SyncMethodWithSyncSuffix = "VSD0041";
```
Vannevelj: Typo: should be `SyncMethodWithAsyncSuffix `

VSDiagnostics/VSDiagnostics/VSDiagnostics/VSDiagnosticsResources.resx
```diff
@@ -486,4 +486,14 @@
   <data name="FlagsEnumValuesAreNotPowersOfTwoValuesDontFitAnalyzerTitle" xml:space="preserve">
     <value>[Flags] enum its values are not explicit powers of 2 and its values dont fit in the specified enum type.</value>
   </data>
+  <data name="SyncMethodWithSyncSuffixAnalyzerMessage" xml:space="preserve">
+    <value>Method "{0}" should not end with 'Async'.</value>
```
Vannevelj: I would specify this as 'Synchronous method (..)' to make it clear why it should be done. You could also update the other analyzer's message to specify the opposite: 'Asynchronous method (..)'

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Async/SyncMethodWithAsyncSuffix/SyncMethodWithAsyncSuffixAnalyzer.cs
tiesmaster: Agreed. I will move it to an extension method tomorrow. The check if a method is async should also be moved to an extension method, imo. Is that ok too?

I think that will simplify the unit tests for the analyzers tremendously, since a lot of scenarios deal with the various situations the extension methods would encounter. That's stuff that should be tested by the unit tests of the extension methods, and not the unit tests of these analyzers, isn't it?

tiesmaster:
I've resolved the minor issues. As mentioned, I'm going to do the extracting tomorrow.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Async/SyncMethodWithAsyncSuffix/SyncMethodWithAsyncSuffixAnalyzer.cs
Vannevelj: > The check if a method is async should also be moved to an extension method, imo. Is that ok too?

Yes, definitely!

I agree that it will simplify the unit tests -- we'd only have to test this once and then a few tests for both analyzers should suffice without having to test for async-ness in the inheritance structure.

tiesmaster:
I wasn't able to finish it this morning, and I'm away for the weekend, so don't expect any updates on this before Monday.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Async/SyncMethodWithAsyncSuffix/SyncMethodWithAsyncSuffixAnalyzer.cs
tiesmaster: Added a separate PR to extract the extension method: #302.

tiesmaster:
@Vannevelj Did the changes, as discussed ;)

Vannevelj:
And it's merged! Thank you for your cooperation throughout the process @tiesmaster

PR #302, opened 2015-11-29
Extract async checking to extension method

tiesmaster:
As discussed [here](https://github.com/Vannevelj/VSDiagnostics/pull/301#discussion_r45457711), this extracts the async checking from the `AsyncMethodWithoutAsyncSuffixAnalyzer`. I though it would be best to do that in a separate PR, to keep things small.

In this PR, I've added an extension method to tell you if a method is async, in the broadest way: either declared async (returning `Task`/`Task<>`), or implemented async (has `async` keyword).

I've called it `IsAsync()`, but that might not be the best name, since there is also an `IsAsync` property on `IMethodSymbol`, but I couldn't find a better name. Please tell me if you have suggestions ;)

### DIFF
--- a/VSDiagnostics/VSDiagnostics/VSDiagnostics.Test/Tests/Utilities/ExtensionsTests.cs
+++ b/VSDiagnostics/VSDiagnostics/VSDiagnostics.Test/Tests/Utilities/ExtensionsTests.cs
@@ -118,15 +118,7 @@ namespace VSDiagnostics.Test.Tests.Utilities
         }
     }";
 
-            var tree = CSharpSyntaxTree.ParseText(source);
-            var mscorlib = MetadataReference.CreateFromFile(typeof(object).Assembly.Location);
-            var compilation = CSharpCompilation.Create("MyCompilation", new[] { tree }, new[] { mscorlib });
-            var semanticModel = compilation.GetSemanticModel(tree);
-
-            var root = tree.GetRoot();
-            var method = root.DescendantNodes().OfType<MethodDeclarationSyntax>().First();
-            var methodSymbol = semanticModel.GetDeclaredSymbol(method);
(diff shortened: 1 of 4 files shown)


VSDiagnostics/VSDiagnostics/VSDiagnostics/Utilities/Extensions.cs
```diff
@@ -131,5 +132,12 @@ public static bool IsAlias(this string type)
 
             return _aliasMapping.ContainsKey(type);
         }
+
+        public static bool IsAsync(this IMethodSymbol methodSymbol)
```
Vannevelj: Could you add some XML documentation to it so we see it pop up in-editor? It's definitely going to cause confusion but I can't immediately come up with a better name so we should at least clarify it through that.

VSDiagnostics/VSDiagnostics/VSDiagnostics.Test/Tests/Utilities/ExtensionsTests.cs
```diff
@@ -99,5 +99,148 @@ void Method(string input)
...
+    namespace ConsoleApplication1
+    {
+        class MyClass
+        {
+            void Method()
+            {
+            }
+    }";
+
+            var tree = CSharpSyntaxTree.ParseText(source);
```
Vannevelj: This piece of code to create a compilation and check the method is often repeated. Can you factor this out to a separate method `CheckAsync(boolean)` or something like that?

tiesmaster:
@Vannevelj Addressed the issues. Could you have another look?

Vannevelj:
Perfect! Thank you for your patience

tiesmaster:
No need to thank me. It's important to keep the code clean, and that's where PR are for ;)

PR #303, opened 2015-12-07
Extract IsDefinedInAncestor to extension method

### LINKED ISSUE #288: NullReferenceException on AsyncMethodWithoutAsyncSuffixAnalyzer
Showed up as an 'info' message with the following stacktrace:

	Message AD0001  Analyzer 'VSDiagnostics.Diagnostics.Async.AsyncMethodWithoutAsyncSuffix.AsyncMethodWithoutAsyncSuffixAnalyzer' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'.Enyim.Caching       1   Active Analyzer 'VSDiagnostics.Diagnostics.Async.AsyncMethodWithoutAsyncSuffix.AsyncMethodWithoutAsyncSuffixAnalyzer' threw the following exception:
	'System.NullReferenceException: Object reference not set to an instance of an object.
	   at VSDiagnostics.Diagnostics.Async.AsyncMethodWithoutAsyncSuffix.AsyncMethodWithoutAsyncSuffixAnalyzer.<>c__DisplayClass10_0.<IsDefinedInAncestor>b__0(ISymbol method)
	   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
	   at VSDiagnostics.Diagnostics.Async.AsyncMethodWithoutAsyncSuffix.AsyncMethodWithoutAsyncSuffixAnalyzer.IsDefinedInAncestor(IMethodSymbol methodSymbol)
	   at VSDiagnostics.Diagnostics.Async.AsyncMethodWithoutAsyncSuffix.AsyncMethodWithoutAsyncSuffixAnalyzer.AnalyzeSyntaxNode(SyntaxNodeAnalysisContext context)
	   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.<>c__DisplayClass38_1`1.<ExecuteSyntaxNodeAction>b__1()
	   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.ExecuteAndCatchIfThrows_NoLock(DiagnosticAnalyzer analyzer, Action analyze)'.

Project: EnyimMemcached https://github.com/enyim/EnyimMemcached
             ServiceStack.Redis https://github.com/ServiceStack/ServiceStack.Redis

tiesmaster:
Similar to #302, pull out `IsDefinedInAncestor()` to an extension method.

Maybe, this also fixes #288, however, I wasn't able to reproduce it. If you have a repro in mind, I can easily add some tests for that.


tiesmaster:
@Vannevelj Build is failing, but it the same restoring error as last time.

tiesmaster:
@Vannevelj Could you have a look at this PR?

VSDiagnostics/VSDiagnostics/VSDiagnostics/Utilities/Extensions.cs
```diff
@@ -144,5 +145,37 @@ public static bool IsAsync(this IMethodSymbol methodSymbol)
...
+            {
+                var interfaceMethods = @interface.GetMembers().Select(containingType.FindImplementationForInterfaceMember);
+                if (interfaceMethods.Any(method => method.Equals(methodSymbol)))
+                {
+                    return true;
+                }
+            }
+
+            var baseType = containingType.BaseType;
+            while (baseType != null)
```
Vannevelj: Can you add a test that tests this code and see if it works with `IsDefinedInAncestor` for `AnotherMethod`? I suspect it might not since `object.BaseType` just returns `object` again -- it's a circular reference.

	public class BaseClass
	{
		public virtual void BaseMethod() { }
	}
	
	public class SubClass
	{
		public void AnotherMethod() { }
	}

tiesmaster: Did you really intent the two classes to be disconnected in your example? Or did you mean something like this:

    public class BaseClass
    {
        public virtual void BaseMethod() { }
    }

    public class SubClass : BaseClass
    {
        public void AnotherMethod() { }
    }

Though, regardless what I do, in both examples the loop ends:

image


Shall, I add a commit with the desired unit test as reference?

tiesmaster: I've pushed the intented commit to a [separate branch](https://github.com/tiesmaster/VSDiagnostics/tree/extra-unit-test-for-pr-303), just in case you don't want it in this PR.

Vannevelj: I'm.. really surprised. We had that as an issue at some point (https://github.com/Vannevelj/VSDiagnostics/blob/develop/VSDiagnostics/VSDiagnostics/VSDiagnostics/Utilities/Extensions.cs#L63) and it had to be fixed by explicitly ending the loop if we're in `object`. More on this in the associated issue: https://github.com/Vannevelj/VSDiagnostics/issues/205

This is something we'll have to look into more closely but that can be done separately. I will create a separate issue for this, no need to add any tests. I will merge this PR since all thoughts I had have been cleared up, thanks for staying with me!

tiesmaster: Indeed, that's very weird... Let me know if I can do anything there to help.

When this PR is merged, I'll update the other PR to make use of this, and the other extension method, and remove all the superfluous tests.

> ... thanks for staying with me!

No probs ;) I had the intention to do some regular contribution to this lib, but I'm quite occupied with some other stuff, so it might take a while before I'll be able to put in some new analyzers :(

Vannevelj: Merged! 

Don't worry about it, we all are ;-) I've had to put it on the backburner for a time as well since I'm interning now but as soon as I've graduated in two months, I can crank it up a notch again. Whatever you can contribute is already highly appreciated

PR #319, opened 2015-12-25
Issue307

Vannevelj:


VSDiagnostics/VSDiagnostics/VSDiagnostics/Utilities/Extensions.cs
```diff
@@ -10,9 +10,10 @@
 
 namespace VSDiagnostics.Utilities
 {
+    // TODO: after null checks, throw argumentnullexceptions instead of returning false
     public static class Extensions
     {
-        private static Dictionary<string, string> _aliasMapping = new Dictionary<string, string>
+        private static readonly Dictionary<string, string> AliasMapping = new Dictionary<string, string>
             {
                 {nameof(Int16), "short"},
```
SLaks: This will fail for VB

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Strings/StringDotFormatWithDifferentAmountOfArguments/StringDotFormatWithDifferentAmountOfArgumentsAnalyzer.cs
```diff
@@ -0,0 +1,150 @@
...
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.RegisterSyntaxNodeAction(AnalyzeNode, SyntaxKind.InvocationExpression);
+        }
+
+        private void AnalyzeNode(SyntaxNodeAnalysisContext context)
+        {
+            var invocation = context.Node as InvocationExpressionSyntax;
```
SLaks: Instead of duplicating all of the work of finding `Format()` invocations, you should put both of these rules in the same analyzer class.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Strings/StringDotFormatWithDifferentAmountOfArguments/StringDotFormatWithDifferentAmountOfArgumentsAnalyzer.cs
```diff
@@ -0,0 +1,150 @@
...
+                return;
+            }
+
+            var hasObjectArray = 
+                methodSymbol.Parameters.FirstOrDefault(x => x.Type.Kind == SymbolKind.ArrayType && 
+                                                            ((IArrayTypeSymbol) x.Type).ElementType.SpecialType == SpecialType.System_Object) != null;
+            var hasObject = 
+                methodSymbol.Parameters.FirstOrDefault(x => x.Type.SpecialType == SpecialType.System_Object) != null;
+
+            if (!(hasObject || hasObjectArray))
```
SLaks: To be safer, you should check that _all_ remaining parameters (`.Skip(formatParam.Ordinal).All(...)`) are either `object` or `params object[]`

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Strings/StringDotFormatWithDifferentAmountOfArguments/StringDotFormatWithDifferentAmountOfArgumentsAnalyzer.cs
```diff
@@ -0,0 +1,150 @@
...
+                var invocationExpression = formatArguments[0].Expression as InvocationExpressionSyntax;
+                if (invocationExpression != null)
+                {
+                    return;
+                }
+
+                // If it's an identifier, we don't handle those that provide an array as a single argument
+                // Other types are fine though -- think about string.Format("{0}", name);
+                var referencedIdentifier = formatArguments[0].Expression as IdentifierNameSyntax;
+                if (referencedIdentifier != null)
```
SLaks: Actually, you should only run this check if the corresponding parameter is `object[]`.

Passing an array as an `object` is fine, and should be treated as a single value.  (although it won't be very useful, since arrays don't override `ToString()`)

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Strings/StringDotFormatWithDifferentAmountOfArguments/StringDotFormatWithDifferentAmountOfArgumentsAnalyzer.cs
```diff
@@ -0,0 +1,150 @@
...
+                // We don't handle method calls
+                var invocationExpression = formatArguments[0].Expression as InvocationExpressionSyntax;
+                if (invocationExpression != null)
+                {
+                    return;
+                }
+
+                // If it's an identifier, we don't handle those that provide an array as a single argument
+                // Other types are fine though -- think about string.Format("{0}", name);
+                var referencedIdentifier = formatArguments[0].Expression as IdentifierNameSyntax;
```
SLaks: Checking for `IdentifierNameSyntax` is not enough; that won't catch parenthesized expressions, method calls, etc...

As mentioned in my next comment, what you really want to check for is a non-array-creation node of array type passed to a parameter of array type

VSDiagnostics/VSDiagnostics/VSDiagnostics/Utilities/Extensions.cs
SLaks: In fact, you should use Roslyn's http://source.roslyn.io/#Microsoft.CodeAnalysis.Workspaces/Simplification/Simplifier.cs,161 instead of all of this.

That will also respect the user's settings.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Strings/StringDotFormatWithDifferentAmountOfArguments/StringDotFormatWithDifferentAmountOfArgumentsAnalyzer.cs
Vannevelj: Yeah, once this analyzer is done I will take everything from this, apply it to the other one and see what I can extract.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Utilities/Extensions.cs
Vannevelj: Ah, I forgot about the Simplifier. I will look into that in a separate PR, thanks.

Vannevelj: Hah. I was going to create a new issue and it looks like I've already came across this once: https://github.com/Vannevelj/VSDiagnostics/issues/269

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Strings/StringDotFormatWithDifferentAmountOfArguments/StringDotFormatWithDifferentAmountOfArgumentsAnalyzer.cs
SLaks: You shouldn't extract anything; instead, put everything into one analyzer.

Otherwise, you'll end up analyzing every node twice, hurting performance.

Vannevelj: Oh like that. Yeah I guess it would make sense since they're both about formatting strings and I've got one or two more that will be related to this.

PS: If you want to earn CR rep, I put out a bounty on this earlier. http://codereview.stackexchange.com/questions/115051/adding-compile-time-format-validation-to-your-string-format-calls

Issue #263, opened 2015-09-23
Implement RedundantXMLDocReturn


Vannevelj:
A `void` method is pointless with a `<return></return>` clause in the XML documentation.

This:

	/// <returns> A number </returns>
	public void MyMethod() { }

Becomes:

	public void MyMethod() { }

Issue #188, opened 2015-08-18
Implement XMLCommentMissingParameter


Vannevelj:
When you have an XML comment like below that is missing a parameter, show a warning. I assume that when someone takes the time to write documentation for one parameter, they generally want that method documented. Forgetting about a second parameter can be for example because it was added afterwards.

        /// <summary>
        ///     Some Summary
        /// </summary>
        /// <param name="o">First parameter</param>
        public static bool MyMethod(object o, int x)

should become

        /// <summary>
        ///     Some Summary
        /// </summary>
        /// <param name="o">First parameter</param>
        /// <param name="x"></param>
        public static bool MyMethod(object o, int x)

PR #345, opened 2016-01-05
Feature262

### LINKED ISSUE #262: Implement RedundantXMLDocParameter
This resembles #188 but works the other way around: show a warning and offer to remove an XML doc parameter comment if there is no corresponding parameter.

This:

	/// <param name="a"> First param </param>
	/// <param name="b"> Non-existing param </param>
	public void MyMethod(int a) { }

becomes

	/// <param name="a"> First param </param>
	public void MyMethod(int a) { }

Hosch250:
Closes #262, #263, & #188

Don't merge this until the current build is released.


Hosch250:
Closing because not all required members are supported.

PR #479, opened 2016-03-12
Close #410

### LINKED ISSUE #410: Expand ConditionIsAlwaysTrue/False to constant expressions
`ConditionIsAlwaysFalse`/ `ConditionIsAlwaysTrue` can both be improved by checking whether the expression itself is a constant value. This would also allow us to check for something like this as well:

    private const bool x = true;
    if(x == true) { }

Instead of restricting us to `if(true)` and `if(false)`.

Hosch250:


VSDiagnostics/VSDiagnostics/VSDiagnostics/VSDiagnosticsResources.resx
```diff
@@ -421,23 +421,14 @@
   <data name="EnumCanHaveFlagsAttributeCodeFixTitle" xml:space="preserve">
     <value>Add Flags attribute</value>
   </data>
-  <data name="ConditionIsAlwaysTrueAnalyzerMessage" xml:space="preserve">
-    <value>Condition is always true.</value>
+  <data name="ConditionIsConstantAnalyzerMessage" xml:space="preserve">
+    <value>Condition is constant.</value>
```
Vannevelj: I would suggest that the message becomes "The condition is always {0}" and as argument you pass in `true`/`false`

VSDiagnostics/VSDiagnostics/VSDiagnostics/VSDiagnosticsResources.resx
```diff
@@ -421,23 +421,14 @@
...
-  <data name="ConditionIsAlwaysFalseAnalyzerMessage" xml:space="preserve">
-    <value>Condition is always false.</value>
-  </data>
-  <data name="ConditionIsAlwaysFalseAnalyzerTitle" xml:space="preserve">
-    <value>Condition is always false.</value>
-  </data>
-  <data name="ConditionIsAlwaysFalseCodeFixTitle" xml:space="preserve">
-    <value>Remove condition and body.</value>
+  <data name="ConditionIsConstantCodeFixTitle" xml:space="preserve">
+    <value>Remove condition and dead code.</value>
```
Vannevelj: I would make the message "Remove condition and unreachable code". "dead code" doesn't tell tell me much about why it's dead.

PR #481, opened 2016-03-13
Analyzer and tests for #4

Hosch250:


VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/LoopedRandomInstantiation/LoopedRandomInstantiationAnalyzer.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        private const DiagnosticSeverity Severity = DiagnosticSeverity.Warning;
+        private static readonly string Category = VSDiagnosticsResources.GeneralCategory;
+        private static readonly string Message = VSDiagnosticsResources.LoopedRandomInstantiationAnalyzerMessage;
+        private static readonly string Title = VSDiagnosticsResources.LoopedRandomInstantiationAnalyzerTitle;
+
+        internal static DiagnosticDescriptor Rule =>
+                new DiagnosticDescriptor(DiagnosticId.LoopedRandomInstantiation, Title, Message, Category, Severity, true);
+
+        private SyntaxKind[] _loopTypes = {SyntaxKind.ForEachStatement, SyntaxKind.ForStatement, SyntaxKind.WhileStatement};
```
Vannevelj: Does this also include the `do..while`? Can you also provide a test for that?

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/LoopedRandomInstantiation/LoopedRandomInstantiationAnalyzer.cs
```diff
@@ -0,0 +1,60 @@
...
+            {
+                return;
+            }
+
+            SyntaxNode currentNode = variableDeclaration;
+            while (!currentNode.IsKind(SyntaxKind.ClassDeclaration))
+            {
+                if (_loopTypes.Contains(currentNode.Kind()))
+                {
+                    context.ReportDiagnostic(Diagnostic.Create(Rule, variableDeclaration.GetLocation()));
```
Vannevelj: Can you create a separate warning for every `VariableDeclarator`? That way we have 2 warnings for for example `var r = new Random(), x = new Random();`.

VSDiagnostics/VSDiagnostics/VSDiagnostics/VSDiagnosticsResources.resx
```diff
@@ -494,4 +494,10 @@
   <data name="StringDotFormatWithDifferentAmountOfArgumentsTitle" xml:space="preserve">
     <value>Invalid arguments for the string.Format() call.</value>
   </data>
+  <data name="LoopedRandomInstantiationAnalyzerMessage" xml:space="preserve">
+    <value>Instantiate the random instance outside the loop.</value>
```
Vannevelj: This doesn't read like something you see in an error list. "*Variable <var> of type Random is instantiated in a loop*" seems more appropriate (note that you have to tell which variabledeclarator it is).

VSDiagnostics/VSDiagnostics/VSDiagnostics/VSDiagnosticsResources.resx
```diff
@@ -494,4 +494,10 @@
...
+  </data>
+  <data name="LoopedRandomInstantiationAnalyzerTitle" xml:space="preserve">
+    <value>Random instance created in loop.</value>
```
Vannevelj: More verbose: "An instance of type System.Random is created in a loop"

VSDiagnostics/VSDiagnostics/VSDiagnostics.Test/Tests/General/LoopedRandomInstantiationTests.cs
```diff
@@ -0,0 +1,172 @@
...
+        {
+            var rand = new Random();
+        }
+    }
+}";
+
+            VerifyDiagnostic(original);
+        }
+    }
+}
```
Vannevelj: Can you add a test that shows we don't just look at the level right above it? 

```csharp
while (true)
{
	if (5 > 6)
	{
		var x = new Random();
	}
}
```

PR #483, opened 2016-03-14
Close #211

### LINKED ISSUE #211: Implement SwitchDoesNotHandleAllEnumOptions
Show a warning when we `switch` over an enum but not all enum options are covered by a `case` statement. I would also add a code fix that inserts the missing ones.

For example:

	void Main()
	{
		var e = MyEnum.A;
		switch(e)
		{
			case A: break;
			case B: break;
			default: break;
		}
	}
	
	enum MyEnum
	{
		A,
		B,
		C
	}

Hosch250:
Closes #211

### DIFF
--- a/VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/SwitchDoesNotHandleAllEnumOptions/SwitchDoesNotHandleAllEnumOptionsCodeFix.cs
+++ b/VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/SwitchDoesNotHandleAllEnumOptions/SwitchDoesNotHandleAllEnumOptionsCodeFix.cs
@@ -1,11 +1,15 @@
-﻿using System.Collections.Immutable;
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Composition;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CodeActions;
 using Microsoft.CodeAnalysis.CodeFixes;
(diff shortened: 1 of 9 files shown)


Vannevelj:
I think you can check if something is an enum by doing `type.SpecialType == SpecialType_Enum` or something like that.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/SwitchDoesNotHandleAllEnumOptions/SwitchDoesNotHandleAllEnumOptionsCodeFix.cs
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        private bool EnumIsUsingStatic(CompilationUnitSyntax root, INamedTypeSymbol enumType)
+        {
+            var fullyQualifiedName = enumType.Name;
+
+            var containingNamespace = enumType.ContainingNamespace;
+            while (!string.IsNullOrEmpty(containingNamespace.Name))
+            {
+                fullyQualifiedName = fullyQualifiedName.Insert(0, containingNamespace.Name + ".");
```
Vannevelj: This is not accurate in the case of nested enums. See for example

```csharp
public class X
{
	public enum Y
	{
		A, B, C
	}
}
```

Which has as fullyqualified name `X+Y`.

The approach seems hacky altogether -- Why do we need to check this in the first place? Can't we just add `Simplifier.Annotation` if 

```csharp
caseLabels.OfType<IdentifierNameSyntax>().Any() || !caseLabels.OfType<MemberAccessExpressionSyntax>().Any()
```

PR #485, opened 2016-03-14
Close #85

### LINKED ISSUE #85: Reference ourselves
I want VSDiagnostics to be an example implementation of VSDiagnostics. Reference the NuGet package and fix all issues we find (I know there are several violations already).

Word of caution: I've noticed that sometimes after adding the VSDiagnostics NuGet package, the solution doesn't build anymore. If this occurs, go to the bottom of it because apparently this should not be happening.

Hosch250:


Hosch250:
I have no idea what happened here.

Hosch250:
I will re-implement this after the other three I committed.  It appears that making everything into expression bodied members broke things.

PR #486, opened 2016-03-14
Remove VB.NET diagnostics.

Hosch250:


Vannevelj:
Can you also remove the Microsoft.CodeAnalysis.VisualBasic dll reference?

Hosch250:
I actually tried, but the tests failed to compile.

Issue #269, opened 2015-10-15
Simplify alias lookup through simplifier


Vannevelj:
I think we're using a lookup table to map explicit types to aliases. We should be able to simplify this using the `Simplifier`.  

https://github.com/dotnet/roslyn/blob/a4290a6f38316641b6aea1f693d19f40bcb7c35e/src/Workspaces/Core/Portable/Simplification/Simplifier.cs

Vannevelj:
That being said -- it's probably a more expensive operation so look into the internals first.

SLaks:
The internals is http://source.roslyn.io/#Microsoft.CodeAnalysis.CSharp.Workspaces/Extensions/ExpressionSyntaxExtensions.cs,2251

Vannevelj:
Done in #487

PR #487, opened 2016-03-15
Close issue #269

Hosch250:


VSDiagnostics/VSDiagnostics/VSDiagnostics.Test/Tests/General/UseAliasesInsteadOfConcreteTypeTests.cs
```diff
@@ -1596,13 +1596,13 @@ class Foo
     {
         static void Main()
         {
-            string bar = ""test"";
+            Single bar = ""test"";
```
Vannevelj: I don't get it. Why is this a fix in the first place? Shouldn't it have no warning?

Hosch250: Right, I need to change that.  Thanks for catching it.

PR #505, opened 2016-03-19
Close #172

### LINKED ISSUE #172: Implement EqualsAndGetHashcodeNotImplementedTogether
When either of the two methods are implemented, the other one has to be as well. I would suggest to add a codefix which implements the method signature and throws a `NotImplementedException()` from the body.

Hosch250:
Cloes #172

### DIFF
--- a/VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/EqualsAndGetHashcodeNotImplementedTogether/EqualsAndGetHashcodeNotImplementedTogetherAnalyzer.cs
+++ b/VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/EqualsAndGetHashcodeNotImplementedTogether/EqualsAndGetHashcodeNotImplementedTogetherAnalyzer.cs
@@ -50,12 +50,18 @@ namespace VSDiagnostics.Diagnostics.General.EqualsAndGetHashcodeNotImplementedTo
                     continue;
                 }
 
-                if (context.SemanticModel.GetDeclaredSymbol(methodDeclaration).OverriddenMethod == objectEquals)
+                var methodSymbol = context.SemanticModel.GetDeclaredSymbol(methodDeclaration).OverriddenMethod;
+                while (methodSymbol.IsOverride)
+                {
+                    methodSymbol = methodSymbol.OverriddenMethod;
+                }
+
+                if (methodSymbol == objectEquals)
                 {
                     equalsImplemented = true;
                 }
 
-                if (context.SemanticModel.GetDeclaredSymbol(methodDeclaration).OverriddenMethod == objectGetHashCode)
+                if (methodSymbol == objectGetHashCode)
                 {
                     getHashcodeImplemented = true;
                 }
--- a/VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/EqualsAndGetHashcodeNotImplementedTogether/EqualsAndGetHashcodeNotImplementedTogetherAnalyzer.cs
+++ b/VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/EqualsAndGetHashcodeNotImplementedTogether/EqualsAndGetHashcodeNotImplementedTogetherAnalyzer.cs
@@ -1,0 +1,72 @@
+﻿using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using VSDiagnostics.Utilities;
+
+namespace VSDiagnostics.Diagnostics.General.EqualsAndGetHashcodeNotImplementedTogether
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class EqualsAndGetHashcodeNotImplemented : DiagnosticAnalyzer
+    {
+        private const DiagnosticSeverity Severity = DiagnosticSeverity.Warning;
+
+        private static readonly string Category = VSDiagnosticsResources.GeneralCategory;
+        private static readonly string Message = VSDiagnosticsResources.EqualsAndGetHashcodeNotImplementedTogetherAnalyzerMessage;
+        private static readonly string Title = VSDiagnosticsResources.EqualsAndGetHashcodeNotImplementedTogetherAnalyzerTitle;
+
+        internal static DiagnosticDescriptor Rule
+            => new DiagnosticDescriptor(DiagnosticId.EqualsAndGetHashcodeNotImplementedTogether, Title, Message, Category, Severity, true);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.RegisterSyntaxNodeAction(AnalyzeNode, SyntaxKind.ClassDeclaration);
+        }
+
+        private void AnalyzeNode(SyntaxNodeAnalysisContext context)
+        {
+            var classDeclaration = (ClassDeclarationSyntax)context.Node;
+
+            var equalsImplemented = false;
+            var getHashcodeImplemented = false;
+
+            foreach (var node in classDeclaration.Members)
+            {
+                if (!node.IsKind(SyntaxKind.MethodDeclaration))
+                {
+                    continue;
+                }
+
+                var methodDeclaration = (MethodDeclarationSyntax)node;
+
+                if (!IsOverride(methodDeclaration))
+                {
+                    continue;
+                }
+
+                var typeInfo = context.SemanticModel.GetTypeInfo(methodDeclaration).Type;   // why isn't this working?
+            }
+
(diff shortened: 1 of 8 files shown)


VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/EqualsAndGetHashcodeNotImplementedTogether/EqualsAndGetHashcodeNotImplementedTogetherAnalyzer.cs
```diff
@@ -0,0 +1,79 @@
...
+                }
+
+                var methodDeclaration = (MethodDeclarationSyntax)node;
+
+                if (!IsOverride(methodDeclaration))
+                {
+                    continue;
+                }
+
+                if (methodDeclaration.Identifier.ValueText == nameof(Equals) && methodDeclaration.ParameterList.Parameters.Count == 1)
```
Vannevelj: I don't like this approach. However there is a more solid alternative: you can request the symbol of `Object+Equals(object)` with something like this:

	var objectSymbol = compilation.GetSpecialType(SpecialType.System_Object);
	var objectEquals = objectSymbol.GetMembers().OfType<IMethodSymbol>().Where(x => x.MetadataName == "Equals" && x.Parameters.Count() == 1);

Afterwards you can use `context.SemanticModel.GetDeclaredSymbol(methodDeclaration).OverriddenMethod` and compare it to the the `Object+Equals(object)` method. 

Some things to note:

* Fetch `Object+Equals(object)` only once per compilation. You can do this by registering a compilation action in `Initialize()`
* You have to iterate through the inheritance structure in case a parent class already overrides it

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/EqualsAndGetHashcodeNotImplementedTogether/EqualsAndGetHashcodeNotImplementedTogetherAnalyzer.cs
```diff
@@ -0,0 +1,79 @@
...
+            }
+
+            if (equalsImplemented ^ getHashcodeImplemented)
+            {
+                context.ReportDiagnostic(Diagnostic.Create(Rule, classDeclaration.GetLocation(),
+                    ImmutableDictionary.CreateRange(new[] { new KeyValuePair<string, string>("IsEqualsImplemented", equalsImplemented.ToString()) })));
+            }
+        }
+
+        private bool IsOverride(MethodDeclarationSyntax methodDeclaration)
```
Vannevelj: This is overzealous with LINQ: a `Contains()` check is perfectly fine (it doesn't allocate anything).

VSDiagnostics/VSDiagnostics/VSDiagnostics/VSDiagnosticsResources.resx
```diff
@@ -509,4 +509,13 @@
   <data name="SwitchDoesNotHandleAllEnumOptionsCodeFixTitle" xml:space="preserve">
     <value>Add case(s).</value>
   </data>
+  <data name="EqualsAndGetHashcodeNotImplementedTogetherAnalyzerMessage" xml:space="preserve">
+    <value>Both Equals() and GetHashcode() must be implemented.</value>
```
Vannevelj: I would reword this to "Equals() and GetHashcode() must be implemented together" to stress that that discrepancy is the problem

VSDiagnostics/VSDiagnostics/VSDiagnostics/VSDiagnosticsResources.resx
```diff
@@ -509,4 +509,13 @@
...
+  </data>
+  <data name="EqualsAndGetHashcodeNotImplementedTogetherAnalyzerTitle" xml:space="preserve">
+    <value>Implement Equals() and GetHashcode().</value>
```
Vannevelj: That's the message for a different analyzer (the one that just provides a context action). Instead, I would make it "Implement Equals() and GetHashcode() together"

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/EqualsAndGetHashcodeNotImplementedTogether/EqualsAndGetHashcodeNotImplementedTogetherCodeFix.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var diagnostic = context.Diagnostics.First();
+            var diagnosticSpan = diagnostic.Location.SourceSpan;
+
+            var statement = root.FindNode(diagnosticSpan);
+
+            if (bool.Parse(diagnostic.Properties["IsEqualsImplemented"]))
```
Vannevelj: You can compress this `if` by moving the entire `context.RegisterCodeFix()` right outside of it and just taking care of the code fix title inside the if (perhaps even by using a ternary if).

Something like this:

	var message = bool.Parse(diagnostic.Properties["IsEqualsImplemented"])
					? "GetHashCode()"
					: "Equals()";

	context.RegisterCodeFix(
		 CodeAction.Create(string.Format(VSDiagnosticsResources.EqualsAndGetHashcodeNotImplementedTogetherCodeFixTitle, message),
		 x => ImplementGetHashCodeAsync(context.Document, root, statement),
		 EqualsAndGetHashcodeNotImplementedTogetherAnalyzer.Rule.Id),
		 diagnostic);

Vannevelj:
Additionally: can you update `README.md` to include the new analyzer?

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/EqualsAndGetHashcodeNotImplementedTogether/EqualsAndGetHashcodeNotImplementedTogetherAnalyzer.cs
Hosch250: This isn't a `Contains` call,  the Linq version would be `return methodDeclaration.Modifiers.Any(modifier => modifier.IsKind(SyntaxKind.OverrideKeyword));`

Hosch250:
I'll update the Readme when it gets merged.  I think there are a couple others missing as well.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/EqualsAndGetHashcodeNotImplementedTogether/EqualsAndGetHashcodeNotImplementedTogetherAnalyzer.cs
```diff
@@ -39,18 +45,17 @@ private void AnalyzeNode(SyntaxNodeAnalysisContext context)
...
 
                 var methodDeclaration = (MethodDeclarationSyntax)node;
-
                 if (!IsOverride(methodDeclaration))
                 {
                     continue;
                 }
 
-                if (methodDeclaration.Identifier.ValueText == nameof(Equals) && methodDeclaration.ParameterList.Parameters.Count == 1)
+                if (context.SemanticModel.GetDeclaredSymbol(methodDeclaration).OverriddenMethod == objectEquals)
```
Vannevelj: What does `OverriddenMethod` refer to? Will this still work if you have a class `X : Object` that implements both methods but also have `Y : X` which violates the rule? Can you add a test to make sure this works?

Hosch250: This case does not work.  For the base class, the containing namespace of the parent Equals/GetHashCode is System.Object.  For the derived class, the parent namespace is the namespace X is defined in.  This causes the expected and actual Equals/GetHashCode methods to be different.

Hosch250: Of course, I could recurse the `OverriddenMethod` value to see if we ever reach the expected override on `Object`.  That should solve the problem.  To answer your question, `OverriddenMethod` refers to the method the current method is overridding.

Vannevelj: Yeah, that was the solution I was alluding to. I think you've already added a commit that fixes this now

Hosch250: Yep, that was #4aec46c

PR #521, opened 2016-04-07
Close #520.  Also remove Linq from this analyzer.

### LINKED ISSUE #520: SwitchDoesNotHandleAllEnumOptions fails when using cast integer
The following code breaks this fix:

    var e = FileMode.Append;
    switch (e)
    {
        case (FileMode)4:
            throw new System.NotImplementedException();
    }

It generates all members, including the item with value `4`, which is uncompilable.

Hosch250:


VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/SwitchDoesNotHandleAllEnumOptions/SwitchDoesNotHandleAllEnumOptionsAnalyzer.cs
```diff
@@ -26,24 +28,63 @@ internal static DiagnosticDescriptor Rule
...
+
+            var labelSymbols = new List<ISymbol>();
+            foreach (var label in caseLabels)
+            {
+                var symbol = context.SemanticModel.GetSymbolInfo(label).Symbol;
+                if (symbol == null)
+                {
+                    // potentially malformed case statement
+                    // or an integer being cast to an enum type
+                    return;
```
Vannevelj: I think you mean to use `continue` here, otherwise you're not going to enumerate the rest of your labels once an invalid one has been found. If that's intended (is it?) then it still doesn't work as it should because the rest of the method will go on with the existing labels that have already been added.

Hosch250: This is the analyzer.  The point is to not create a diagnostic if any of the case statements cannot be analyzed correctly with `GetSymbolInfo`.

Vannevelj: Oh right. The `return` statement also breaks out of the encompassing method. Alright, all good then I think.

PR #532, opened 2016-04-11
Remove Linq from Analyzers

### LINKED ISSUE #493: Avoid LINQ in analyzers
LINQ is an allocation-expensive operation that is not recommended for use in analyzers.

Hosch250:
I did not remove it from the extensions or handlers (yet).  I didn't keep any `Contains` or `Any` calls that use lambdas, and I didn't keep any `Select`, `Where`, or `All` calls.  I also created a couple handy extensions methods to help with some calls, such as a `ContainsAny` that checks whether the items in a list contain any of the items in another list and a `Contains` to check whether a list of `SyntaxNode`s contain any of a given kind.

Closes #493

### DIFF
--- a/VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Attributes/FlagsEnumValuesAreNotPowersOfTwo/FlagsEnumValuesAreNotPowersOfTwoAnalyzer.cs
+++ b/VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Attributes/FlagsEnumValuesAreNotPowersOfTwo/FlagsEnumValuesAreNotPowersOfTwoAnalyzer.cs
@@ -79,7 +79,7 @@ namespace VSDiagnostics.Diagnostics.Attributes.FlagsEnumValuesAreNotPowersOfTwo
 
             var enumName = context.SemanticModel.GetDeclaredSymbol(declarationExpression).Name;
             var enumMemberDeclarations =
-                declarationExpression.ChildNodes().SyntaxNodeOfType<EnumMemberDeclarationSyntax>(SyntaxKind.EnumMemberDeclaration);
+                declarationExpression.ChildNodes().OfType<EnumMemberDeclarationSyntax>(SyntaxKind.EnumMemberDeclaration);
 
             foreach (var member in enumMemberDeclarations)
             {
--- a/VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Attributes/FlagsEnumValuesAreNotPowersOfTwo/FlagsEnumValuesAreNotPowersOfTwoAnalyzer.cs
+++ b/VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Attributes/FlagsEnumValuesAreNotPowersOfTwo/FlagsEnumValuesAreNotPowersOfTwoAnalyzer.cs
@@ -1,12 +1,12 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
-using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using VSDiagnostics.Utilities;
+// ReSharper disable LoopCanBeConvertedToQuery
 
 namespace VSDiagnostics.Diagnostics.Attributes.FlagsEnumValuesAreNotPowersOfTwo
 {
@@ -51,14 +51,26 @@ namespace VSDiagnostics.Diagnostics.Attributes.FlagsEnumValuesAreNotPowersOfTwo
         private void AnalyzeSymbol(SyntaxNodeAnalysisContext context)
         {
             var declarationExpression = (EnumDeclarationSyntax) context.Node;
-            var flagsAttribute = declarationExpression.AttributeLists.FirstOrDefault(
-                a => a.Attributes.FirstOrDefault(
-                    t =>
-                    {
-                        var symbol = context.SemanticModel.GetSymbolInfo(t).Symbol;
-                        return symbol == null || symbol.ContainingType.MetadataName == typeof(FlagsAttribute).Name;
-                    }) != null);
 
+            AttributeListSyntax flagsAttribute = null;
+
+            foreach (var list in declarationExpression.AttributeLists)
+            {
+                foreach (var attribute in list.Attributes)
+                {
+                    var symbol = context.SemanticModel.GetSymbolInfo(attribute).Symbol;
+                    if (symbol == null || symbol.ContainingType.MetadataName == typeof (FlagsAttribute).Name)
+                    {
+                        flagsAttribute = list;
+                        break;
+                    }
+                }
+
+                if (flagsAttribute != null)
+                {
+                    break; 
+                }
+            }
 
             if (flagsAttribute == null)
             {
@@ -67,7 +79,7 @@ namespace VSDiagnostics.Diagnostics.Attributes.FlagsEnumValuesAreNotPowersOfTwo
 
             var enumName = context.SemanticModel.GetDeclaredSymbol(declarationExpression).Name;
             var enumMemberDeclarations =
-                declarationExpression.ChildNodes().OfType<EnumMemberDeclarationSyntax>().ToList();
+                declarationExpression.ChildNodes().OfType<EnumMemberDeclarationSyntax>(SyntaxKind.EnumMemberDeclaration);
 
             foreach (var member in enumMemberDeclarations)
             {
@@ -76,11 +88,26 @@ namespace VSDiagnostics.Diagnostics.Attributes.FlagsEnumValuesAreNotPowersOfTwo
                     continue;
                 }
 
-                var descendantNodes = member.EqualsValue.Value.DescendantNodesAndSelf().ToList();
-                if (descendantNodes.OfType<LiteralExpressionSyntax>().Any() &&
-                    descendantNodes.OfType<IdentifierNameSyntax>().Any())
+                var descendantNodes = member.EqualsValue.Value.DescendantNodesAndSelf();
+
+                var containsLiteralExpression = false;
+                var containsIdentifierName = false;
+                foreach (var node in descendantNodes)
                 {
-                    return;
+                    if (node is LiteralExpressionSyntax)
+                    {
+                        containsLiteralExpression = true;
+                    }
+
+                    if (node.IsKind(SyntaxKind.IdentifierName))
+                    {
+                        containsIdentifierName = true;
+                    }
+
+                    if (containsIdentifierName && containsLiteralExpression)
+                    {
+                        return;
+                    }
                 }
             }
 
@@ -126,9 +153,19 @@ namespace VSDiagnostics.Diagnostics.Attributes.FlagsEnumValuesAreNotPowersOfTwo
 
                 if (member.EqualsValue.Value is BinaryExpressionSyntax)
                 {
-                    var descendantNodes = member.EqualsValue.Value.DescendantNodesAndSelf().ToList();
-                    if (descendantNodes.Any() &&
-                        descendantNodes.All(n => n is IdentifierNameSyntax || n is BinaryExpressionSyntax))
+                    var descendantNodes = new List<SyntaxNode>(member.EqualsValue.Value.DescendantNodesAndSelf());
+
+                    var all = true;
+                    foreach (var node in descendantNodes)
+                    {
+                        if (!node.IsKind(SyntaxKind.IdentifierName) && !(node is BinaryExpressionSyntax))
+                        {
+                            all = false;
+                            break;
+                        }
+                    }
+
+                    if (descendantNodes.Any() && all)
                     {
                         continue;
                     }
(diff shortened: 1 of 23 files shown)


VSDiagnostics/VSDiagnostics/VSDiagnostics/Utilities/Extensions.cs
```diff
@@ -205,5 +206,62 @@ public static bool IsNameofInvocation(this InvocationExpressionSyntax invocation
 
             return identifier != null && identifier.Identifier.ValueText == "nameof";
         }
+
+        public static List<T> SyntaxNodeOfType<T>(this IEnumerable<SyntaxNode> enumerable, SyntaxKind kind) where T : SyntaxNode
```
Vannevelj: This one isn't necessary since `OfType<T>` doesn't allocated a lambda anyway and I believe this implementation is largely the same.

Vannevelj: Let's stick with the same name `OfType` -- method overloading should not be distinguished based on name.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Utilities/Extensions.cs
```diff
@@ -205,5 +206,62 @@ public static bool IsNameofInvocation(this InvocationExpressionSyntax invocation
...
+                if (node.IsKind(kind))
+                {
+                    list.Add((T)node);
+                }
+            }
+
+            return list;
+        }
+
+        public static bool ContainsAny(this SyntaxTokenList list, SyntaxKind[] kinds)
```
Vannevelj: can we make this a `params`? That should make it a little easier to call

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Attributes/FlagsEnumValuesAreNotPowersOfTwo/FlagsEnumValuesAreNotPowersOfTwoAnalyzer.cs
```diff
@@ -67,7 +79,7 @@ private void AnalyzeSymbol(SyntaxNodeAnalysisContext context)
 
             var enumName = context.SemanticModel.GetDeclaredSymbol(declarationExpression).Name;
             var enumMemberDeclarations =
-                declarationExpression.ChildNodes().OfType<EnumMemberDeclarationSyntax>().ToList();
+                declarationExpression.ChildNodes().SyntaxNodeOfType<EnumMemberDeclarationSyntax>(SyntaxKind.EnumMemberDeclaration);
```
Vannevelj: If it's going to be called with an argument (`SyntaxKind`) then it shouldn't be called with a type `T`. I think you can just make the method non-generic

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Exceptions/ArgumentExceptionWithoutNameofOperator/ArgumentExceptionWithoutNameofOperatorAnalyzer.cs
```diff
@@ -41,12 +42,20 @@ private void AnalyzeSyntaxNode(SyntaxNodeAnalysisContext context)
...
             if (symbolInformation.Symbol.InheritsFrom(typeof(ArgumentException)))
             {
-                var arguments =
-                    objectCreationExpression.ArgumentList.Arguments.Select(x => x.Expression)
-                                            .OfType<LiteralExpressionSyntax>();
+                var arguments = new List<LiteralExpressionSyntax>();
+
+                foreach (var argument in objectCreationExpression.ArgumentList.Arguments)
+                {
+                    var expression = argument.Expression as LiteralExpressionSyntax;
```
Vannevelj: use is + direct cast

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/NullableToShorthand/NullableToShorthandAnalyzer.cs
```diff
@@ -64,17 +64,28 @@ private void AnalyzeSymbol(SyntaxNodeAnalysisContext context)
...
+
+                    if (parentNode == null)
+                    {
+                        return;
+                    }
                 }
 
                 if (parentNode.Kind() == SyntaxKind.LocalDeclarationStatement)
                 {
+                    // ReSharper disable once PossibleInvalidCastException
```
Vannevelj: let's not add all these resharper annotations. I don't like library-specific things in the codebase unless we can't avoid it.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/SwitchDoesNotHandleAllEnumOptions/SwitchDoesNotHandleAllEnumOptionsAnalyzer.cs
```diff
@@ -81,8 +81,8 @@ private void AnalyzeSymbol(SyntaxNodeAnalysisContext context)
                 }
 
                 if (!switchHasSymbol)
-                {
-                    context.ReportDiagnostic(Diagnostic.Create(Rule, switchBlock.GetLocation()));
+            {
```
Vannevelj: wrong indentation?

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/TryCastWithoutUsingAsNotNull/TryCastWithoutUsingAsNotNullAnalyzer.cs
```diff
@@ -99,7 +99,10 @@ private void AnalyzeSymbol(SyntaxNodeAnalysisContext context)
                 }
             };
 
-            foreach (var expression in asExpressions.Concat<ExpressionSyntax>(castExpressions))
+            var expressions = new List<ExpressionSyntax>(asExpressions);
```
Vannevelj: Initialize the list with `new List<ExpressionSyntax(asExpressions.Length + castExpressions.Length)` and then use `AddRange` twice. This will prevent a resizing of the internal list because it already had the correct size from the start

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Strings/StringDotFormatWithDifferentAmountOfArguments/StringDotFormatWithDifferentAmountOfArgumentsAnalyzer.cs
```diff
@@ -37,28 +39,62 @@ private void AnalyzeNode(SyntaxNodeAnalysisContext context)
             var invokedMethod = context.SemanticModel.GetSymbolInfo(invocation);
```
Vannevelj: is `invokedMethod` even used anywhere except the line below? Because if it's not, you can just remove this intermediate variable

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Strings/StringDotFormatWithDifferentAmountOfArguments/StringDotFormatWithDifferentAmountOfArgumentsAnalyzer.cs
```diff
@@ -102,14 +143,23 @@ private void AnalyzeNode(SyntaxNodeAnalysisContext context)
...
                     {
                         // We don't handle method calls that return an array in the case of a single argument
                         return;
                     }
 
-                    var inlineArrayCreation = formatArguments[0].DescendantNodes().OfType<InitializerExpressionSyntax>().FirstOrDefault();
+                    InitializerExpressionSyntax inlineArrayCreation = null;
+                    foreach (var argument in formatArguments[0].DescendantNodes())
+                    {
+                        var argumentAsInitializerExpressionSyntax = argument as InitializerExpressionSyntax;
```
Vannevelj: is + direct cast

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Strings/StringDotFormatWithDifferentAmountOfArguments/StringDotFormatWithDifferentAmountOfArgumentsAnalyzer.cs
```diff
@@ -131,19 +181,20 @@ private void AnalyzeNode(SyntaxNodeAnalysisContext context)
...
+                placeholders.Add(int.Parse(PlaceholderHelpers.GetPlaceholderIndex(placeholder.Value)));
+            }
 
             if (!placeholders.Any())
             {
                 return;
             }
 
-            var highestPlaceholder = placeholders.Max();
+            var highestPlaceholder = placeholders.Concat(new[] {int.MinValue}).Max();
```
Vannevelj: What does this change do? What effect should it have?

Hosch250: I've no idea what this change does.  It looks like it ensures there is a minimum value.  R# did this when I converted the loop back.

Vannevelj: It should be removed since I don't see any benefit to it. It will always have a value -- there's no point in adding that min-value.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Utilities/Extensions.cs
Hosch250: This might result in an ambiguous call that cannot be resolved.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Attributes/FlagsEnumValuesAreNotPowersOfTwo/FlagsEnumValuesAreNotPowersOfTwoAnalyzer.cs
Hosch250: I'm pretty sure I made it generic so I would get an explicit type back, rather than just `SyntaxNode`.  If I change it, I'm going to need a bunch of casts in the code.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Utilities/Extensions.cs
Vannevelj: Why not? The other overload doesn't have a `SyntaxKind` argument. Can you verify it?

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Attributes/FlagsEnumValuesAreNotPowersOfTwo/FlagsEnumValuesAreNotPowersOfTwoAnalyzer.cs
Vannevelj: Fair enough

VSDiagnostics/VSDiagnostics/VSDiagnostics/Utilities/Extensions.cs
```diff
@@ -154,21 +186,26 @@ public static bool IsDefinedInAncestor(this IMethodSymbol methodSymbol)
...
             {
                 var interfaceMethods =
-                    @interface.GetMembers().Select(containingType.FindImplementationForInterfaceMember).Where(x => x != null);
+                    @interface.GetMembers().Select(containingType.FindImplementationForInterfaceMember).ToList();
 
-                if (interfaceMethods.Any(method => method.Equals(methodSymbol)))
+                foreach (var interfaceMethod in interfaceMethods)
                 {
-                    return true;
+                    if (interfaceMethod != null && interfaceMethod.Equals(methodSymbol))
```
Vannevelj: can we use null propagation here? I'm not entirely sure

VSDiagnostics/VSDiagnostics/VSDiagnostics/Utilities/Extensions.cs
```diff
@@ -154,21 +186,26 @@ public static bool IsDefinedInAncestor(this IMethodSymbol methodSymbol)
...
+                    }
                 }
             }
 
             var baseType = containingType.BaseType;
             while (baseType != null)
             {
-                var baseMethods = baseType.GetMembers().OfType<IMethodSymbol>();
-                if (baseMethods.Any(method => method.Equals(methodSymbol.OverriddenMethod)))
+                foreach (var baseMethod in baseType.GetMembers().OfType<IMethodSymbol>())
```
Vannevelj: Wasn't there an extension to replace `OfType<>`? Or doesn't that work with symbols?

VSDiagnostics/VSDiagnostics/VSDiagnostics/Utilities/Extensions.cs
Hosch250: It creates a `bool?`, so I have to check for `null` anyway.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Utilities/Extensions.cs
Hosch250: `IMethodSymbol` is used for any "method-like symbol", including ctors and properties.  My custom `OfType` works only for known SyntaxKind types.

Issue #236, opened 2015-08-23
Implement IntegerDivisionByInteger


Vannevelj:
When dividing two integers, the fraction is truncated from the result effectively resulting in a rounding. Figure out what the equivalent is in a more expressive way and provide a warning + code fix.

I imagine it might turn from this:

    int x = 5;
    int y = 6;
    double z = 5 / 6;

to this:

    int x = 5;
    int y = 6;
    double z = Math.Floor((double) 5 / (double) 6);

Completely untested and I might be making things up, but you get the idea.

Hosch250:
This will need lot of test cases:

```
int x = 5 / 6;
var x = 5 / 6;
double x = 5 / 6;
var x = (double) (5 / 6);
var x = (int) (5 / 6);
var x = 5 / 6 - 3;
var x = 3 + 5 / 6;
var x = 3 == 5 / 6;

void Foo (int x) {}
Foo (5 / 6);

void Foo (double x) {}
Foo (5 / 6);
```

Any others?

Vannevelj:
Actually, I don't think we need the code fix as laid out in the first post since it is most likely a hidden problem and not intended to work like this. Just casting one of the arguments to a double should suffice as fix.

PR #536, opened 2016-04-22
Implemented #236

Vannevelj:
Implements issue #236

### DIFF
--- a/VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Arithmetic/DivideIntegerByInteger/DivideIntegerByIntegerAnalyzer.cs
+++ b/VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Arithmetic/DivideIntegerByInteger/DivideIntegerByIntegerAnalyzer.cs
@@ -1,0 +1,60 @@
+﻿using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using VSDiagnostics.Utilities;
+
+namespace VSDiagnostics.Diagnostics.Arithmetic.DivideIntegerByInteger
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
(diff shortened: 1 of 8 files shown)


VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Arithmetic/DivideIntegerByInteger/DivideIntegerByIntegerAnalyzer.cs
```diff
@@ -0,0 +1,60 @@
...
+        private const DiagnosticSeverity Severity = DiagnosticSeverity.Warning;
+
+        private static readonly string Category = VSDiagnosticsResources.ArithmeticCategory;
+        private static readonly string Message = VSDiagnosticsResources.DivideIntegerByIntegerAnalyzerMessage;
+        private static readonly string Title = VSDiagnosticsResources.DivideIntegerByIntegerAnalyzerTitle;
+
+        internal static DiagnosticDescriptor Rule
+            =>
+                new DiagnosticDescriptor(DiagnosticId.DivideIntegerByInteger, Title, Message, Category, Severity,
+                    isEnabledByDefault: true);
```
Hosch250: If I understand correctly, diagnostics are enabled by default.  We probably only need argument when we do not want them enabled by default, which would make the difference more clear when skimming the files.

Vannevelj: Good point. I'll remove it from here manually and create an issue for us to remove it everywhere else as well.

Vannevelj: Oh nevermind. It's not an optional argument -- it's just a named argument. I did this here (and everywhere else) because I don't want a boolean argument since it doesn't say anything about what it does.

Hosch250: OK.  When I was working on Roslyn, this argument was optional; perhaps it has been changed in later versions than we are working off of.

PR #542, opened 2016-04-29
Implement #150

### LINKED ISSUE #150: Implement SwitchStatementWithoutDefaultBlock
`switch` statements their `default` block is optional but I believe it should always be included, at the very least to indicate in what scenario none of the `case` blocks would have been hit.

I would suggest to add a code fix that adds the following block (or a similar message):

    default: // This should or shouldn't get hit because..

Hosch250:
Close #150

### DIFF
--- a/VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/SwitchIsMissingDefaultLabel/SwitchIsMissingDefaultLabelAnalyzer.cs
+++ b/VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/SwitchIsMissingDefaultLabel/SwitchIsMissingDefaultLabelAnalyzer.cs
@@ -1,0 +1,49 @@
+﻿using System.Collections.Generic;
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using VSDiagnostics.Utilities;
+
+namespace VSDiagnostics.Diagnostics.General.SwitchIsMissingDefaultLabel
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    internal class SwitchIsMissingDefaultLabelAnalyzer : DiagnosticAnalyzer
+    {
+        private const DiagnosticSeverity Severity = DiagnosticSeverity.Warning;
+
+        private static readonly string Category = VSDiagnosticsResources.GeneralCategory;
+        private static readonly string Message = VSDiagnosticsResources.SwitchIsMissingDefaultSectionAnalyzerMessage;
+        private static readonly string Title = VSDiagnosticsResources.SwitchIsMissingDefaultSectionAnalyzerTitle;
+
+        internal static DiagnosticDescriptor Rule
+            => new DiagnosticDescriptor(DiagnosticId.SwitchIsMissingDefaultLabel, Title, Message, Category, Severity, true);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context) => context.RegisterSyntaxNodeAction(AnalyzeSymbol, SyntaxKind.SwitchStatement);
+
+        private void AnalyzeSymbol(SyntaxNodeAnalysisContext context)
+        {
+            var switchBlock = (SwitchStatementSyntax)context.Node;
+
+            var hasDefaultLabel = false;
+            foreach (var section in switchBlock.Sections)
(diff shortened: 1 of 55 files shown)


VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/SwitchIsMissingDefaultLabel/SwitchIsMissingDefaultLabelAnalyzer.cs
```diff
@@ -0,0 +1,49 @@
...
+        internal static DiagnosticDescriptor Rule
+            => new DiagnosticDescriptor(DiagnosticId.SwitchIsMissingDefaultLabel, Title, Message, Category, Severity, true);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context) => context.RegisterSyntaxNodeAction(AnalyzeSymbol, SyntaxKind.SwitchStatement);
+
+        private void AnalyzeSymbol(SyntaxNodeAnalysisContext context)
+        {
+            var switchBlock = (SwitchStatementSyntax)context.Node;
```
Vannevelj: This will cause an exception when you have a parenthesized switch like this:

```csharp
	var x = 5;

	switch ((x))
	{
		case 5: 
		case 6:
		default:
			break;
	}
```

Add this as a test. You can probably just look through the ancestors to find the first of `SyntaxKind.SwitchStatement`.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/SwitchIsMissingDefaultLabel/SwitchIsMissingDefaultLabelCodeFix.cs
```diff
@@ -0,0 +1,57 @@
...
+            var statement = root.FindNode(diagnosticSpan);
+                context.RegisterCodeFix(
+                    CodeAction.Create(VSDiagnosticsResources.SwitchIsMissingDefaultSectionCodeFixTitle,
+                        x => AddDefaultCaseAsync(context.Document, (CompilationUnitSyntax) root, statement),
+                        SwitchIsMissingDefaultLabelAnalyzer.Rule.Id), diagnostic);
+        }
+
+        private async Task<Solution> AddDefaultCaseAsync(Document document, CompilationUnitSyntax root, SyntaxNode statement)
+        {
+            var switchBlock = (SwitchStatementSyntax)statement.Parent;
```
Vannevelj: This will cause an exception when you have a parenthesized switch like this:

```csharp
	var x = 5;

	switch ((x))
	{
		case 5: 
		case 6:
		default:
			break;
	}
```

Add this as a test. You can probably just look through the ancestors to find the first of `SyntaxKind.SwitchStatement`.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/SwitchIsMissingDefaultLabel/SwitchIsMissingDefaultLabelCodeFix.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            var notImplementedException =
+                SyntaxFactory.ThrowStatement(SyntaxFactory.ParseExpression(" new System.NotImplementedException()"))
```
Vannevelj: Why is there a space here? Is it because it otherwise puts `throw` and `new` together? If so: just add `Formatter.Annotation` as well and remove the space.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/SwitchIsMissingDefaultLabel/SwitchIsMissingDefaultLabelCodeFix.cs
```diff
@@ -0,0 +1,57 @@
...
+            var statements = SyntaxFactory.List(new List<StatementSyntax> { notImplementedException });
+
+            var defaultCase = SyntaxFactory.SwitchSection(SyntaxFactory.List<SwitchLabelSyntax>(new[] {SyntaxFactory.DefaultSwitchLabel()}), statements);
+
+            var newNode = switchBlock.AddSections(defaultCase)
+                .WithAdditionalAnnotations(Formatter.Annotation, Simplifier.Annotation);
+
+            var newRoot = root.ReplaceNode(switchBlock, newNode);
+            var newDocument = await Simplifier.ReduceAsync(document.WithSyntaxRoot(newRoot));
+            return newDocument.Project.Solution;
```
Vannevelj: Go for the `Task<Document>` return value since you're not touching anything else in the project anyway.

PR #543, opened 2016-04-29
Close #150

### LINKED ISSUE #150: Implement SwitchStatementWithoutDefaultBlock

Hosch250:


VSDiagnostics/VSDiagnostics/VSDiagnostics/Utilities/DiagnosticId.cs
```diff
@@ -48,5 +48,6 @@ public static class DiagnosticId
         public const string DivideIntegerByInteger = "VSD0045";
         public const string EqualsAndGetHashcodeNotImplementedTogether = "VSD0046";
         public const string ElementaryMethodsOfTypeInCollectionNotOverridden = "VSD0047";
+        public const string SwitchIsMissingDefaultLabel = "VSD0049";
```
Vannevelj: Where did VSD0048 go to? Do you still need to merge develop into your branch?

Vannevelj: Oh nevermind. I didn't merge the PR myself yet

PR #548, opened 2016-04-29
Change exception

Hosch250:


VSDiagnostics/VSDiagnostics/VSDiagnostics.Test/Tests/General/SwitchIsMissingDefaultLabelTests.cs
```diff
@@ -57,7 +57,7 @@ void Method()
                 case MyEnum.Buzz:
                     break;
                 default:
-                    throw new System.NotImplementedException();
+                    throw new System.ArgumentException();
```
Vannevelj: These should have an argument. `throw new System.ArgumentException(nameof(e))`

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/SwitchIsMissingDefaultLabel/SwitchIsMissingDefaultLabelCodeFix.cs
```diff
@@ -37,15 +37,31 @@ public override ImmutableArray<string> FixableDiagnosticIds
...
-                SyntaxFactory.ThrowStatement(SyntaxFactory.ParseExpression("new System.NotImplementedException()"))
+            var model = await document.GetSemanticModelAsync();
+
+            var symbol = model.GetSymbolInfo(switchBlock.Expression).Symbol;
+            string expression;
+            if (symbol != null)
+            {
+                expression = $"nameof({symbol.Name})";
+            }
+            else if (switchBlock.Expression.IsKind(SyntaxKind.StringLiteralExpression))
```
Vannevelj: Instead of the elseif/if, can't you just do `else { expression = switchBlock.Expression.ToString(); }`?

Hosch250: That results in `("test".ToString())`, even with the simplifier annotation.

Vannevelj: Oh okay, I see now what you were doing. Okay, that should suffice as it is then.

PR #550, opened 2016-04-29
Feature185

### LINKED ISSUE #185: Implement StructWithoutElementaryMethodsOverridden
A `struct`'s default implementation for `GetHashCode()`, `Equals(object)` and `.ToString()` heavily uses reflection. A Code Fix that provides a default implementation would be good here:

	struct MyStruct
	{
		public override int GetHashCode()
		{
			throw new NotImplementedException();
		}
		
		public override bool Equals(object o)
		{
			throw new NotImplementedException();
		}
		
		public override string ToString()
		{
			throw new NotImplementedException();
		}
	}

Hosch250:
Close #185

### DIFF
--- a/VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Attributes/OnPropertyChangedWithoutCallerMemberName/OnPropertyChangedWithoutCallerMemberNameAnalyzer.cs
+++ b/VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Attributes/OnPropertyChangedWithoutCallerMemberName/OnPropertyChangedWithoutCallerMemberNameAnalyzer.cs
@@ -34,6 +34,11 @@ namespace VSDiagnostics.Diagnostics.Attributes.OnPropertyChangedWithoutCallerMem
         {
             var methodDeclaration = (MethodDeclarationSyntax)context.Node;
             var parentClass = methodDeclaration.Ancestors().OfType<ClassDeclarationSyntax>(SyntaxKind.ClassDeclaration).FirstOrDefault();
+            if (parentClass == null)
+            {
+                return;
+            }
+
             var typeSymbol = context.SemanticModel.GetDeclaredSymbol(parentClass);
 
             // class must implement INotifyPropertyChanged
--- a/VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Attributes/OnPropertyChangedWithoutCallerMemberName/OnPropertyChangedWithoutCallerMemberNameAnalyzer.cs
+++ b/VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Attributes/OnPropertyChangedWithoutCallerMemberName/OnPropertyChangedWithoutCallerMemberNameAnalyzer.cs
@@ -34,10 +34,6 @@ namespace VSDiagnostics.Diagnostics.Attributes.OnPropertyChangedWithoutCallerMem
         {
             var methodDeclaration = (MethodDeclarationSyntax)context.Node;
             var parentClass = methodDeclaration.Ancestors().OfType<ClassDeclarationSyntax>(SyntaxKind.ClassDeclaration).FirstOrDefault();
-            if (parentClass == null)
-            {
-                return;
-            }
 
             var typeSymbol = context.SemanticModel.GetDeclaredSymbol(parentClass);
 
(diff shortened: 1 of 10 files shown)


VSDiagnostics/VSDiagnostics/VSDiagnostics/VSDiagnosticsResources.resx
```diff
@@ -543,4 +543,13 @@
   <data name="SwitchIsMissingDefaultSectionCodeFixTitle" xml:space="preserve">
     <value>Add default label</value>
   </data>
+  <data name="StructWithoutElementaryMethodsOverriddenAnalyzerMessage" xml:space="preserve">
+    <value>Structs should implement Equals(), GetHashCode() and ToString().</value>
+  </data>
+  <data name="StructWithoutElementaryMethodsOverriddenAnalyzerTitle" xml:space="preserve">
+    <value>Implement elementary methods</value>
```
Vannevelj: Specify which type it refers to: "Implement elementary methods on struct {0}"

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Structs/StructWithoutElementaryMethodsOverridden/StructWithoutElementaryMethodsOverriddenCodeFix.cs
```diff
@@ -0,0 +1,113 @@
...
+            if (!implementToString)
+            {
+                newStatement = newStatement.AddMembers(GetToStringMethod());
+            }
+
+            var newRoot = root.ReplaceNode(statement, newStatement);
+            return Task.FromResult(document.WithSyntaxRoot(newRoot));
+        }
+
+        private MethodDeclarationSyntax GetEqualsMethod()
```
Vannevelj: Can you make three static fields of type `MethodDeclarationSyntax` instead and use the static constructor of `StructWithoutElementaryMethodsOverriddenCodeFix` to assign these fields? That way they will only be calculated once (when the class is loaded) and every time the CodeFix is executed it can just return the `MethodDeclarationSyntax` it computed.

Hosch250: OK.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Attributes/OnPropertyChangedWithoutCallerMemberName/OnPropertyChangedWithoutCallerMemberNameAnalyzer.cs
```diff
@@ -34,6 +34,11 @@ private void AnalyzeSymbol(SyntaxNodeAnalysisContext context)
         {
             var methodDeclaration = (MethodDeclarationSyntax)context.Node;
             var parentClass = methodDeclaration.Ancestors().OfType<ClassDeclarationSyntax>(SyntaxKind.ClassDeclaration).FirstOrDefault();
+            if (parentClass == null)
+            {
+                return;
+            }
+
```
Hosch250: I had a null reference exception here.  This fixed it, but I'll file an issue so we can inspect it more thoroughly.

Vannevelj:
branch conflict

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Structs/StructWithoutElementaryMethodsOverridden/StructWithoutElementaryMethodsOverriddenCodeFix.cs
```diff
@@ -0,0 +1,177 @@
...
+            if (!implementToString)
+            {
+                newStatement = newStatement.AddMembers(ToStringMethod);
+            }
+
+            var newRoot = root.ReplaceNode(statement, newStatement);
+            return Task.FromResult(document.WithSyntaxRoot(newRoot));
+        }
+
+        private static MethodDeclarationSyntax GetEqualsMethod()
```
Vannevelj: These still need to be moved to static fields. I'm not sure if that message got lost earlier

PR #551, opened 2016-04-29
Close #544

### LINKED ISSUE #544: Loop statement without braces shows wrong message
Something slipped through the cracks. `LoopStatementWithoutBracesAnalyzer` its resources use what should be `LoopedRandomInstantiationAnalyzer` its resources while the latter uses an outdated version.

Hosch250:
Close #544


Vannevelj:
You should also change `LoopedRandomInstantiationAnalyzer` its messages to what this one incorrectly had (e.g. "Variable {0} of type System.Random is instantiated in a loop." and "An instance of type System.Random is created in a loop.").

PR #552, opened 2016-04-29
Feature545

### LINKED ISSUE #545: Implement UsingStatementWithoutBraces
A `using()` statement should be followed with braces as well. An exception to this is when you have multiple chained using statements: in that case only the last one should have the braces.

Hosch250:
Close #545.

Do not merge this until after #551 is merged.


Vannevelj:
Can you put this in a separate branch? You're integrating #544 again.

Don't forget the README as well

PR #556, opened 2016-04-30
Implement #477

### LINKED ISSUE #477: Implement ImplementEqualsAndHashcode
Provide a basic implementation. This should work as a context action when the caret is on the class declaration. Related to #185 and #3

https://msdn.microsoft.com/en-us/library/336aedhh(v=vs.71).aspx

Hosch250:
Close #477

### DIFF
--- a/VSDiagnostics/VSDiagnostics/VSDiagnostics.Test/Tests/General/ImplementEqualsAndGetHashCodeTests.cs
+++ b/VSDiagnostics/VSDiagnostics/VSDiagnostics.Test/Tests/General/ImplementEqualsAndGetHashCodeTests.cs
@@ -129,12 +129,18 @@ namespace ConsoleApplication1
             }
 
             var value = (MyClass)obj;
-            return _foo == value._foo && _bar == value._bar && Foo == value.Foo && Bar == value.Bar;
+            return _foo == value._foo &&
+                   _bar == value._bar &&
+                   Foo == value.Foo &&
+                   Bar == value.Bar;
         }
 
         public override int GetHashCode()
         {
-            return _foo.GetHashCode() ^ _bar.GetHashCode() ^ Foo.GetHashCode() ^ Bar.GetHashCode();
+            return _foo.GetHashCode() ^
+                   _bar.GetHashCode() ^
+                   Foo.GetHashCode() ^
+                   Bar.GetHashCode();
         }
     }
 }";
@@ -184,12 +190,16 @@ namespace ConsoleApplication1
             }
 
             var value = (MyClass)obj;
-            return _foo == value._foo && _bar == value._bar && Foo == value.Foo;
+            return _foo == value._foo &&
+                   _bar == value._bar &&
+                   Foo == value.Foo;
         }
 
         public override int GetHashCode()
         {
-            return _foo.GetHashCode() ^ _bar.GetHashCode() ^ Foo.GetHashCode();
+            return _foo.GetHashCode() ^
+                   _bar.GetHashCode() ^
+                   Foo.GetHashCode();
         }
     }
 }";
@@ -355,12 +365,18 @@ namespace ConsoleApplication1
             }
 
             var value = (MyStruct)obj;
-            return _foo == value._foo && _bar == value._bar && Foo == value.Foo && Bar == value.Bar;
+            return _foo == value._foo &&
+                   _bar == value._bar &&
+                   Foo == value.Foo &&
+                   Bar == value.Bar;
         }
 
         public override int GetHashCode()
         {
-            return _foo.GetHashCode() ^ _bar.GetHashCode() ^ Foo.GetHashCode() ^ Bar.GetHashCode();
+            return _foo.GetHashCode() ^
+                   _bar.GetHashCode() ^
+                   Foo.GetHashCode() ^
+                   Bar.GetHashCode();
         }
     }
 }";
@@ -410,12 +426,16 @@ namespace ConsoleApplication1
             }
 
             var value = (MyStruct)obj;
-            return _foo == value._foo && _bar == value._bar && Foo == value.Foo;
+            return _foo == value._foo &&
+                   _bar == value._bar &&
+                   Foo == value.Foo;
         }
 
         public override int GetHashCode()
         {
-            return _foo.GetHashCode() ^ _bar.GetHashCode() ^ Foo.GetHashCode();
+            return _foo.GetHashCode() ^
+                   _bar.GetHashCode() ^
+                   Foo.GetHashCode();
         }
     }
 }";
@@ -491,12 +511,14 @@ namespace ConsoleApplication1
             }
 
             var value = (MyClass)obj;
-            return _foo == value._foo && _bar == value._bar;
+            return _foo == value._foo &&
+                   _bar == value._bar;
         }
 
         public override int GetHashCode()
         {
-            return _foo.GetHashCode() ^ _bar.GetHashCode();
+            return _foo.GetHashCode() ^
+                   _bar.GetHashCode();
         }
     }
 }";
--- a/VSDiagnostics/VSDiagnostics/VSDiagnostics.Test/Tests/General/ImplementEqualsAndGetHashCodeTests.cs
+++ b/VSDiagnostics/VSDiagnostics/VSDiagnostics.Test/Tests/General/ImplementEqualsAndGetHashCodeTests.cs
@@ -213,7 +213,7 @@ namespace ConsoleApplication1
         }
 
         [TestMethod]
-        public void ImplementEqualsAndGetHashCode_ImplementsEquals_HasField()
+        public void ImplementEqualsAndGetHashCode_ClassImplementsEquals_HasField()
         {
             var original = @"
 namespace ConsoleApplication1
@@ -236,6 +236,316 @@ namespace ConsoleApplication1
 }";
 
             VerifyDiagnostic(original);
+        }
+
+        [TestMethod]
+        public void ImplementEqualsAndGetHashCode_StructDoesNotImplementEither_HasField()
+        {
+            var original = @"
+namespace ConsoleApplication1
+{
+    struct MyStruct
+    {
+        string _foo;
+    }
+}";
+
+            var result = @"
+namespace ConsoleApplication1
+{
+    struct MyStruct
+    {
+        string _foo;
+
+        public override bool Equals(object obj)
+        {
+            if (obj == null || typeof(MyStruct) != obj.GetType())
+            {
+                return false;
+            }
+
+            var value = (MyStruct)obj;
+            return _foo == value._foo;
+        }
+
+        public override int GetHashCode()
+        {
+            return _foo.GetHashCode();
+        }
+    }
+}";
+
+            VerifyDiagnostic(original, string.Format(ImplementEqualsAndGetHashCodeAnalyzer.Rule.MessageFormat.ToString(), "MyStruct"));
+            VerifyFix(original, result);
+        }
+
+        [TestMethod]
+        public void ImplementEqualsAndGetHashCode_StructDoesNotImplementEither_HasProperty()
+        {
+            var original = @"
+namespace ConsoleApplication1
+{
+    struct MyStruct
+    {
+        string Foo { get; set; }
+    }
+}";
+
+            var result = @"
+namespace ConsoleApplication1
+{
+    struct MyStruct
+    {
+        string Foo { get; set; }
+
+        public override bool Equals(object obj)
+        {
+            if (obj == null || typeof(MyStruct) != obj.GetType())
+            {
+                return false;
+            }
+
+            var value = (MyStruct)obj;
+            return Foo == value.Foo;
+        }
+
+        public override int GetHashCode()
+        {
+            return Foo.GetHashCode();
+        }
+    }
+}";
+
+            VerifyDiagnostic(original, string.Format(ImplementEqualsAndGetHashCodeAnalyzer.Rule.MessageFormat.ToString(), "MyStruct"));
+            VerifyFix(original, result);
+        }
+
+        [TestMethod]
+        public void ImplementEqualsAndGetHashCode_StructDoesNotImplementEither_HasMultipleFieldsAndProperties()
+        {
+            var original = @"
+namespace ConsoleApplication1
+{
+    struct MyStruct
+    {
+        string _foo;
+        string _bar;
+
+        string Foo { get; set; }
+        string Bar { get; set; }
+    }
+}";
+
+            var result = @"
+namespace ConsoleApplication1
+{
+    struct MyStruct
+    {
+        string _foo;
+        string _bar;
+
+        string Foo { get; set; }
+        string Bar { get; set; }
+
+        public override bool Equals(object obj)
+        {
+            if (obj == null || typeof(MyStruct) != obj.GetType())
+            {
+                return false;
+            }
+
+            var value = (MyStruct)obj;
+            return _foo == value._foo && _bar == value._bar && Foo == value.Foo && Bar == value.Bar;
+        }
+
+        public override int GetHashCode()
+        {
+            return _foo.GetHashCode() ^ _bar.GetHashCode() ^ Foo.GetHashCode() ^ Bar.GetHashCode();
+        }
+    }
+}";
+
+            VerifyDiagnostic(original, string.Format(ImplementEqualsAndGetHashCodeAnalyzer.Rule.MessageFormat.ToString(), "MyStruct"));
+            VerifyFix(original, result);
+        }
+
+        [TestMethod]
+        public void ImplementEqualsAndGetHashCode_StructDoesNotImplementEither_HasSetOnlyProperty()
+        {
+            var original = @"
+namespace ConsoleApplication1
+{
+    struct MyStruct
+    {
+        string _foo;
+        string _bar;
+
+        string Foo { get; set; }
+        string Bar
+        {
+            set { _bar = value; }
+        }
+    }
+}";
+
+            var result = @"
+namespace ConsoleApplication1
+{
+    struct MyStruct
+    {
+        string _foo;
+        string _bar;
+
+        string Foo { get; set; }
+        string Bar
+        {
+            set { _bar = value; }
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (obj == null || typeof(MyStruct) != obj.GetType())
+            {
+                return false;
+            }
+
+            var value = (MyStruct)obj;
+            return _foo == value._foo && _bar == value._bar && Foo == value.Foo;
+        }
+
+        public override int GetHashCode()
+        {
+            return _foo.GetHashCode() ^ _bar.GetHashCode() ^ Foo.GetHashCode();
+        }
+    }
+}";
+
+            VerifyDiagnostic(original, string.Format(ImplementEqualsAndGetHashCodeAnalyzer.Rule.MessageFormat.ToString(), "MyStruct"));
+            VerifyFix(original, result);
+        }
+
+        [TestMethod]
+        public void ImplementEqualsAndGetHashCode_StructDoesNotImplementEither_DoesNotHaveFieldOrProperty()
+        {
+            var original = @"
+namespace ConsoleApplication1
+{
+    struct MyStruct
+    {
+    }
+}";
+
+            VerifyDiagnostic(original);
+        }
+
+        [TestMethod]
+        public void ImplementEqualsAndGetHashCode_StructImplementsEquals_HasField()
+        {
+            var original = @"
+namespace ConsoleApplication1
+{
+    struct MyStruct
+    {
+        string _foo;
+
+        public override bool Equals(object obj)
+        {
+            if (obj == null || typeof(MyStruct) != obj.GetType())
+            {
+                return false;
+            }
+
+            var value = (MyStruct)obj;
+            return _foo == value._foo;
+        }
+    }
+}";
+
+            VerifyDiagnostic(original);
+        }
+
+        [TestMethod]
+        public void ImplementEqualsAndGetHashCode_ClassDoesNotImplementEither_HasChainedFields()
+        {
+            var original = @"
+namespace ConsoleApplication1
+{
+    class MyClass
+    {
+        string _foo = ""test"", _bar = ""test"";
+    }
+}";
+
+            var result = @"
+namespace ConsoleApplication1
+{
+    class MyClass
+    {
+        string _foo = ""test"", _bar = ""test"";
+
+        public override bool Equals(object obj)
+        {
+            if (obj == null || typeof(MyClass) != obj.GetType())
+            {
+                return false;
+            }
+
+            var value = (MyClass)obj;
+            return _foo == value._foo && _bar == value._bar;
+        }
+
+        public override int GetHashCode()
+        {
+            return _foo.GetHashCode() ^ _bar.GetHashCode();
+        }
+    }
+}";
+
+            VerifyDiagnostic(original, string.Format(ImplementEqualsAndGetHashCodeAnalyzer.Rule.MessageFormat.ToString(), "MyClass"));
+            VerifyFix(original, result);
+        }
+
+        [TestMethod]
+        public void ImplementEqualsAndGetHashCode_ClassDoesNotImplementEither_HasStaticFields()
+        {
+            var original = @"
+namespace ConsoleApplication1
+{
+    class MyClass
+    {
+        string _foo = ""test"";
+        static string _bar = ""test"";
+    }
+}";
+
+            var result = @"
+namespace ConsoleApplication1
+{
+    class MyClass
+    {
+        string _foo = ""test"";
+        static string _bar = ""test"";
+
+        public override bool Equals(object obj)
+        {
+            if (obj == null || typeof(MyClass) != obj.GetType())
+            {
+                return false;
+            }
+
+            var value = (MyClass)obj;
+            return _foo == value._foo;
+        }
+
+        public override int GetHashCode()
+        {
+            return _foo.GetHashCode();
+        }
+    }
+}";
+
+            VerifyDiagnostic(original, string.Format(ImplementEqualsAndGetHashCodeAnalyzer.Rule.MessageFormat.ToString(), "MyClass"));
+            VerifyFix(original, result);
         }
     }
 }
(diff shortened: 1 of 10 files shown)


VSDiagnostics/VSDiagnostics/VSDiagnostics.Test/Tests/General/ImplementEqualsAndGetHashCodeTests.cs
```diff
@@ -0,0 +1,241 @@
...
+    class MyClass
+    {
+    }
+}";
+
+            VerifyDiagnostic(original);
+        }
+
+        [TestMethod]
+        public void ImplementEqualsAndGetHashCode_ImplementsEquals_HasField()
```
Vannevelj: I think it would be good to still offer the code fix when one of the two methods has already been implemented. This will need a separate message of course

Hosch250: I would agree. This makes [#172](https://github.com/VSDiagnostics/VSDiagnostics/issues/172) somewhat moot then.

Vannevelj: Oh you're right, that one already provides a code fix already. Then we can leave it like this: if there is one of the methods implemented, the other one will be triggered by #172 and be fixed with a NotImplementedException. If none are implemented, we can provide a default implementation through the hidden code action.

Hosch250: I'd almost rather just disable that one and make two rules for this one--one as an error and one as Hidden.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeCodeFix.cs
```diff
@@ -0,0 +1,128 @@
...
+            var castStatement = SyntaxFactory.ParseStatement(
+                    $"var value = ({classDeclaration.Identifier}) obj;{Environment.NewLine}");
+
+            var fieldAndPropertyEqualityStatements = new List<string>();
+            foreach (var member in classDeclaration.Members)
+            {
+                if (member.IsKind(SyntaxKind.FieldDeclaration))
+                {
+                    var field = (FieldDeclarationSyntax)member;
+                    fieldAndPropertyEqualityStatements.AddRange(field.Declaration.Variables.Select(
```
Vannevelj: Can you add a unit test that uses chained field declarators? `private int _field1 = 5, _field2 = 10;` just to have this logic under test

Vannevelj:
This has merge conflicts now. Can you change the analyzer ID to a not-used code? I forgot to account for this in the other analyzers. PS: there will have to be some more changes to this analyzer but we'll look at that soon enough after the Programmers.se question

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeAnalyzer.cs
```diff
@@ -0,0 +1,152 @@
...
+
+                if (MembersDoNotContainOverridenEqualsAndGetHashCode(context.SemanticModel, structDeclaration.Members, objectEquals, objectGetHashCode) &&
+                    MembersContainNonStaticFieldOrProperty(structDeclaration.Members))
+                {
+                    context.ReportDiagnostic(Diagnostic.Create(Rule, structDeclaration.Identifier.GetLocation(), structDeclaration.Identifier));
+                }
+            }
+        }
+
+        private bool MembersDoNotContainOverridenEqualsAndGetHashCode(SemanticModel model, SyntaxList<SyntaxNode> members, IMethodSymbol objectEquals, IMethodSymbol objectGetHashCode)
```
Vannevelj: Don't use negatives -- that just makes everything more confusing

Vannevelj: Additionally: instead of passing in the `SemanticModel` you should just pass a `IMethodSymbol` directly

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeAnalyzer.cs
```diff
@@ -0,0 +1,152 @@
...
+                {
+                    objectGetHashCode = method;
+                }
+            }
+
+            if (context.Node is ClassDeclarationSyntax)
+            {
+                var classDeclaration = (ClassDeclarationSyntax) context.Node;
+
+                if (MembersDoNotContainOverridenEqualsAndGetHashCode(context.SemanticModel, classDeclaration.Members, objectEquals, objectGetHashCode) &&
```
Vannevelj: Instead of passing `classDeclaration.Members` you can just pass `classSymbol.GetMembers("GetHashCode", "Equals")`. Same for structs

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeCodeFix.cs
```diff
@@ -0,0 +1,256 @@
...
+                diagnostic);
+        }
+
+        private async Task<Document> ImplementEqualsAndGetHashCodeAsync(Document document, SyntaxNode root, SyntaxNode statement)
+        {
+            var model = await document.GetSemanticModelAsync();
+
+            var objectSymbol = model.Compilation.GetSpecialType(SpecialType.System_Object);
+            var objectEquals = objectSymbol.GetMembers().OfType<IMethodSymbol>()
+                    .First(method => method.MetadataName == nameof(Equals) && method.Parameters.Count() == 1);
```
Vannevelj: This can be made more semantically correct by using `.Single()`.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeCodeFix.cs
```diff
@@ -0,0 +1,256 @@
...
+
+            var parameter = SyntaxFactory.Parameter(SyntaxFactory.Identifier("obj"))
+                .WithType(SyntaxFactory.ParseTypeName("object"));
+
+            var ifStatement =
+                SyntaxFactory.IfStatement(SyntaxFactory.ParseExpression($"obj == null || typeof({identifier}) != obj.GetType()"),
+                    SyntaxFactory.Block(SyntaxFactory.ParseStatement("return false;")));
+
+            var castStatement = SyntaxFactory.ParseStatement(
+                    $"var value = ({identifier}) obj;{Environment.NewLine}");
```
Vannevelj: Why does this need an explicit newline? Doesn't it get laid out properly otherwise?

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeCodeFix.cs
```diff
@@ -0,0 +1,256 @@
...
+                    if (field.Modifiers.Contains(SyntaxKind.StaticKeyword))
+                    {
+                        continue;
+                    }
+
+                    var symbol = model.GetTypeInfo(field.Declaration.Type).Type;
+                    if (symbol.IsValueType)
+                    {
+                        fieldAndPropertyEqualityStatements.AddRange(field.Declaration.Variables.Select(
+                            variable => $"{variable.Identifier}.Equals(value.{variable.Identifier})"));
```
Vannevelj: Why are structs specifically compared using `.Equals()` and not `==`?

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeCodeFix.cs
```diff
@@ -0,0 +1,256 @@
...
+// https://msdn.microsoft.com/en-us/library/system.object.gethashcode%28v=vs.110%29.aspx
+"));
+
+            return SyntaxFactory.MethodDeclaration(SyntaxFactory.ParseTypeName("int"), "GetHashCode")
+                    .AddModifiers(publicModifier, overrideModifier)
+                    .AddBodyStatements(returnStatement)
+                    .WithAdditionalAnnotations(Formatter.Annotation);
+        }
+
+        private bool BaseClassImplementsEquals(IMethodSymbol objectEquals, INamedTypeSymbol symbol)
```
Vannevelj: This has to iterate through all classes above this one. See this for example:

```csharp
class A
{
	public virtual void M() { }
}

class B : A { }

class C : B
{
	public override void M() { }
}
```

VSDiagnostics/VSDiagnostics/VSDiagnostics/VSDiagnosticsResources.resx
```diff
@@ -571,6 +571,15 @@
   <data name="UsingStatementWithoutBracesCodeFixTitle" xml:space="preserve">
     <value>Add braces</value>
   </data>
+  <data name="ImplementEqualsAndGetHashCodeAnalyzerMessage" xml:space="preserve">
+    <value>Class {0} does not implement Equals() and GetHashCode().</value>
```
Vannevelj: Could also be a `struct`

VSDiagnostics/VSDiagnostics/VSDiagnostics/VSDiagnosticsResources.resx
```diff
@@ -571,6 +571,15 @@
...
+  </data>
+  <data name="ImplementEqualsAndGetHashCodeAnalyzerTitle" xml:space="preserve">
+    <value>Implement Equals() and GetHashCode().</value>
+  </data>
+  <data name="ImplementEqualsAndGetHashCodeCodeFixTitle" xml:space="preserve">
+    <value>Add Equals() and GetHashCode()</value>
```
Vannevelj: > Add `Equals()` and `GetHashCode()` implementations

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeCodeFix.cs
Hosch250: Structs do not implement `==` automatically, leading to a compile error: http://chat.stackexchange.com/transcript/message/29428905#29428905

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeAnalyzer.cs
Hosch250: I need to calculate the method symbol on all overridden methods in the class, so passing in a single `IMethodSymbol` won't work.

Vannevelj: Ah I see that now. Looked past it -- alright.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeAnalyzer.cs
Hosch250: `GetMembers()` doesn't take two string value.  It takes 0 or 1.

Vannevelj: Ah bummer. I thought I saw an overload that took multiple arguments -- nevermind then.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeCodeFix.cs
Hosch250: That is right.  Is there a better way?

VSDiagnostics/VSDiagnostics/VSDiagnostics/VSDiagnosticsResources.resx
Hosch250: Should just change it to `Member`, or should I explicitly say "Class" or "Struct" as applicable?

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeCodeFix.cs
Vannevelj: I'm surprised this statement requires an explicit newline and others statements like the `return false` from above don't. Can you look into why this is necessary and add it as a comment?

VSDiagnostics/VSDiagnostics/VSDiagnostics/VSDiagnosticsResources.resx
Vannevelj: Explicitly mention "Class" or "Struct"

VSDiagnostics/VSDiagnostics/VSDiagnostics.Test/Tests/General/ImplementEqualsAndGetHashCodeTests.cs
```diff
@@ -1313,8 +1373,41 @@ public override int GetHashCode()
     }
 }";
 
-            VerifyDiagnostic(original, string.Format(ImplementEqualsAndGetHashCodeAnalyzer.Rule.MessageFormat.ToString(), "MyClass"));
+            VerifyDiagnostic(original, string.Format(ImplementEqualsAndGetHashCodeAnalyzer.Rule.MessageFormat.ToString(), "Class", "MyClass"));
             VerifyFix(original, result);
         }
+
+        [TestMethod]
+        public void ImplementEqualsAndGetHashCode_ExplicitlyCheckMessage_Class()
```
Vannevelj: This shouldn't be a separate test -- every preceding test just has to do it explicitly.

Hosch250: OK, I'll change that tonight and fix the merge conflict.

Vannevelj:
Can you also add a test where a field uses the `const` keyword? I think right now we only have a test for `static`.

Hosch250:
Good catch.

Vannevelj:
oh lol, I didn't even realize that `Equals()` ignores static (and now const) fields. I was referring to `GetHashCode()`: afaik there is no test that verifies we don't use `const` there.

Hosch250:
I just added one.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeAnalyzer.cs
```diff
@@ -0,0 +1,144 @@
...
+        private static readonly string Category = VSDiagnosticsResources.GeneralCategory;
+        private static readonly string Message = VSDiagnosticsResources.ImplementEqualsAndGetHashCodeAnalyzerMessage;
+        private static readonly string Title = VSDiagnosticsResources.ImplementEqualsAndGetHashCodeAnalyzerTitle;
+
+        internal static DiagnosticDescriptor Rule =>
+            new DiagnosticDescriptor(DiagnosticId.ImplementEqualsAndGetHashCode, Title, Message, Category, Severity, true);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context) =>
```
Vannevelj: There should also be a test in the case of a `partial class` where one file has the declarations (or split/none). In this case we should regard all partial files as a whole. I believe we can avoid a lot of work to support this by doing `context.RegisterSymbolAction(SymbolKind.Class)` though. At that point we have our type symbol already and we can just look at the members inside of it (all partial files will be 'merged' already)

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeCodeFix.cs
```diff
@@ -0,0 +1,274 @@
...
+
+            if (property.AccessorList.Accessors.Any(a => a.IsKind(SyntaxKind.SetAccessorDeclaration)))
+            {
+                return string.Empty;
+            }
+
+            // ensure getter does not have body
+            // the property has to have at least one of {get, set}, and it doesn't have a set (see above)
+            // this will not have an NRE in First()
+            if (property.AccessorList.Accessors.First(a => a.IsKind(SyntaxKind.GetAccessorDeclaration)).Body == null)
```
Vannevelj: Where do we enforce that a property has at least a getter? What if we have an expression-bodied property like 

    public int Prop => 5;

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeAnalyzer.cs
Hosch250: Good catch.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeCodeFix.cs
Hosch250: That is considered a get-only property.  I re-wrote this, but basically, it ensures the IPropertySymbol `GetMethod` property is not null.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeAnalyzer.cs
```diff
@@ -23,38 +21,44 @@ public class ImplementEqualsAndGetHashCodeAnalyzer : DiagnosticAnalyzer
...
-            if (!TypeMembersContainOverridenEqualsOrGetHashCode(context.SemanticModel, declaration.Members, objectEquals, objectGetHashCode) &&
-                MembersContainNonStaticFieldOrProperty(declaration.Members))
+            if (objectEquals != null || objectGetHashCode != null) { return; }
+
+            if (MembersContainNonStaticFieldOrProperty(namedType.GetMembers()))
             {
-                context.ReportDiagnostic(Diagnostic.Create(Rule, declaration.Identifier.GetLocation(),
-                    declaration.IsKind(SyntaxKind.ClassDeclaration) ? "Class" : "Struct", declaration.Identifier));
+                // it is only reported at the main location--the other locations are (apparently) just passed as data
+                // todo--ask the Roslyn team about this
```
Vannevelj: Can you track this in an issue?

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeAnalyzer.cs
```diff
@@ -23,38 +21,44 @@ public class ImplementEqualsAndGetHashCodeAnalyzer : DiagnosticAnalyzer
...
+                for (var i = 0; i < namedType.Locations.Count(); i++)
+                {
+                    symbol.ReportDiagnostic(Diagnostic.Create(Rule, namedType.Locations[i],
+                    namedType.Locations.RemoveAt(i),
```
Vannevelj: What does this do exactly? What's happening here?

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeAnalyzer.cs
```diff
@@ -67,72 +71,27 @@ private void GetEqualsAndGetHashCodeSymbols(SemanticModel model, out IMethodSymb
             }
         }
 
-        private bool TypeMembersContainOverridenEqualsOrGetHashCode(SemanticModel model, SyntaxList<SyntaxNode> members, IMethodSymbol objectEquals, IMethodSymbol objectGetHashCode)
+        private bool MembersContainNonStaticFieldOrProperty(ImmutableArray<ISymbol> members)
```
Vannevelj: A lot of code dissapeared from these helper methods. Why did this become obsolete?

VSDiagnostics/VSDiagnostics/VSDiagnostics.Test/Tests/General/ImplementEqualsAndGetHashCodeTests.cs
```diff
@@ -1421,5 +1421,136 @@ public override int GetHashCode()
...
+}";
+
+            VerifyDiagnostic(original, "Class MyClass does not implement Equals() and GetHashCode().");
+            VerifyFix(original, result);
+        }
+
+        [TestMethod]
+        public void ImplementEqualsAndGetHashCode_PartialClassDoesNotImplementEither_ImplementsAllTypesInSplitClass()
+        {
+            var original = @"
```
Vannevelj: These should be put in separate trees instead of one syntax tree with multiple classes. For example: https://github.com/VSDiagnostics/VSDiagnostics/blob/develop/VSDiagnostics/VSDiagnostics/VSDiagnostics.Test/Tests/General/RedudantPrivateSetterTests.cs#L226

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeAnalyzer.cs
Hosch250: Because it suddenly became much simpler to check the stuff with the properties in `ISymbol` and `IPropertySymbol`.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeAnalyzer.cs
Hosch250: This reports the diagnostic at each location of the declaration.

Vannevelj: So there's a warning for every partial declaration? I'm not entirely sure if that's what we want. What do you think about just picking the first location?

Hosch250: That way, it is only available in one location, which is potentially a `.g.cs` or other generated file.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/ImplementEqualsAndGetHashCode/ImplementEqualsAndGetHashCodeAnalyzer.cs
Hosch250: I just asked the Roslyn team: https://gitter.im/dotnet/roslyn?at=574b8fe080352f204df34446.  Still want an issue?

Vannevelj: No that's fine -- we'll wait to see what they say

Issue #562, opened 2016-05-03
Aggregate: Exception thrown from prohibited context


Vannevelj:
- [x] ExceptionThrownFromImplicitOperator #168
- [x] PropertyGetterThrowsException #423
- [x] StaticConstructorThrowsException #427
- [x] ExceptionThrownFromExceptionFilterBlock #447
- [x] ExceptionThrownFromFinallyBlock #448
- [x] DoNotThrowExceptionFromEqualityOperator #466
- [x] DoNotThrowExceptionFromDispose #472
- [x] DoNotThrowExceptionFromFinalizer  #474
- [x] DoNotThrowExceptionFromGetHashCode  #560
- [x] DoNotThrowExceptionFromEquals #566

PR #565, opened 2016-05-04
implemented #562

Vannevelj:


VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/Exceptions/ExceptionThrownFromProhibitedContext/ExceptionThrownFromProhibitedContextAnalyzer.cs
```diff
@@ -0,0 +1,159 @@
...
+                    var methodName = method.Identifier.ValueText;
+
+                    if (methodName == "Dispose")
+                    {
+                        var arity = method.ParameterList.Parameters.Count;
+                        context.ReportDiagnostic(Diagnostic.Create(DisposeRule, warningLocation, arity == 0 ? "Dispose()" : "Dispose(bool)", containingType.Name));
+                        return;
+                    }
+
+                    if (methodName == "GetHashCode")
```
Hosch250: You could use the same technique I use to get these methods--by looking up the members of the type and comparing them.  That might be somewhat expensive for this analyzer, though, seeing how much it covers.

Vannevelj: Yeah, it seems like a lot of extra work for a very tiny amount of cases (who hides GetHashCode or Equals..). Then again I prefer to have the analyzer entirely correct so I'll change it

PR #601, opened 2016-06-19
GetHashCode() refers to mutable member

### LINKED ISSUE #559: Implement GetHashCodeRefersToMutableField
see https://msdn.microsoft.com/en-us/library/system.object.gethashcode%28v=vs.110%29.aspx

> you should override GetHashCode only if 
> - You can compute the hash code from fields that are not mutable

This includes fields and properties. A mutable field is any field or property that is a reference type. Basically: we can accept any value type.
We can also augment this by allowing every `readonly` class that is deemed to be immutable. We don't do `const` fields because they have -- by definition -- the same value across all instances.

Hosch250:
Close #559


VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/GetHashCodeRefersToMutableMember/GetHashCodeRefersToMutableMemberAnalyzer.cs
```diff
@@ -0,0 +1,182 @@
...
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+        
+        public override void Initialize(AnalysisContext context) =>
+            context.RegisterSymbolAction(AnalyzeSymbol, SymbolKind.NamedType);
+
+        private void AnalyzeSymbol(SymbolAnalysisContext context)
+        {
+            var namedType = (INamedTypeSymbol)context.Symbol;
+            var semanticModel = context.Compilation.GetSemanticModel(context.Symbol.Locations[0].SourceTree);
```
Vannevelj: context.SemanticModel

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/GetHashCodeRefersToMutableMember/GetHashCodeRefersToMutableMemberAnalyzer.cs
```diff
@@ -0,0 +1,182 @@
...
+                    if (propertyIsMutable.Item1)
+                    {
+                        context.ReportDiagnostic(Diagnostic.Create(Rule, getHashCode.Locations[0],
+                            propertyIsMutable.Item2));
+                    }
+                }
+            }
+        }
+
+        private void GetHashCodeSymbol(INamedTypeSymbol symbol, out IMethodSymbol getHashCodeSymbol)
```
Vannevelj: This can just return IMethodSymbol and get rid of the `out `param.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/GetHashCodeRefersToMutableMember/GetHashCodeRefersToMutableMemberAnalyzer.cs
```diff
@@ -0,0 +1,182 @@
...
+        {
+            var namedType = (INamedTypeSymbol)context.Symbol;
+            var semanticModel = context.Compilation.GetSemanticModel(context.Symbol.Locations[0].SourceTree);
+
+            IMethodSymbol getHashCode;
+            GetHashCodeSymbol(namedType, out getHashCode);
+            if (getHashCode == null) { return; }
+
+            var getHashCodeLocation = getHashCode.Locations.FirstOrDefault();
+            var root = getHashCodeLocation?.SourceTree.GetRoot(context.CancellationToken);
```
Vannevelj: `context.Node.SyntaxTree.GetRootAsync()`

VSDiagnostics/VSDiagnostics/VSDiagnostics/VSDiagnosticsResources.resx
```diff
@@ -654,4 +654,10 @@
   <data name="TestMethodWithoutTestAttributeAnalyzerTitle" xml:space="preserve">
     <value>A method might be missing a test attribute.</value>
   </data>
+  <data name="GetHashCodeRefersToMutableFieldAnalyzerMessage" xml:space="preserve">
+    <value>GetHashCode() refers to {0}</value>
```
Vannevelj: Message should probably be changed to "GetHashCode() refers to {0} {1} {2}" with mutable/static/constant, field/property and name.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/GetHashCodeRefersToMutableMember/GetHashCodeRefersToMutableMemberAnalyzer.cs
```diff
@@ -0,0 +1,182 @@
...
+
+                var method = (IMethodSymbol)member;
+                if (method.MetadataName == nameof(GetHashCode) && !method.Parameters.Any())
+                {
+                    getHashCodeSymbol = method;
+                }
+            }
+        }
+        
+        private Tuple<bool, string> FieldIsMutableOrStatic(IFieldSymbol field)
```
Vannevelj: It feels more natural to have `Tuple<string, bool>` since the key is the property/fieldname

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/GetHashCodeRefersToMutableMember/GetHashCodeRefersToMutableMemberAnalyzer.cs
```diff
@@ -0,0 +1,182 @@
...
+
+            if (property.SetMethod != null)
+            {
+                description += "settable ";
+                returnResult = true;
+            }
+
+            var propertyLocation = property.Locations.FirstOrDefault();
+            var root = propertyLocation?.SourceTree.GetRoot(token);
+            if (root == null)
```
Vannevelj: We've calculated the `root` before, you can just pass that in again

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/GetHashCodeRefersToMutableMember/GetHashCodeRefersToMutableMemberAnalyzer.cs
```diff
@@ -0,0 +1,182 @@
...
+            }
+
+            var propertyNode = (PropertyDeclarationSyntax) root.FindNode(propertyLocation.SourceSpan);
+
+            // ensure getter does not have body
+            // the property has to have at least one of {get, set}, and it doesn't have a set (see above)
+            // this will not have an NRE in First()
+            // the accessor list might be null if it uses the arrow operator `=>`
+            if (propertyNode.AccessorList == null ||
+                propertyNode.AccessorList.Accessors.First(a => a.IsKind(SyntaxKind.GetAccessorDeclaration)).Body != null)
```
Vannevelj: Use `.Single()` instead

VSDiagnostics/VSDiagnostics/VSDiagnostics/VSDiagnosticsResources.resx
Hosch250: I thought about that, but a bodied-getter isn't exactly mutable, but it isn't allowed either.

VSDiagnostics/VSDiagnostics/VSDiagnostics/Diagnostics/General/GetHashCodeRefersToMutableMember/GetHashCodeRefersToMutableMemberAnalyzer.cs
Hosch250: Actually, I can use `[0]` and get rid of the Linq completely.

VSDiagnostics/VSDiagnostics/VSDiagnostics/VSDiagnosticsResources.resx
Vannevelj: It should still be a somewhat different message: "GetHashcode() refers to <blub>" doesn't make sense from an error message perspective. Even "GetHashcode() refers to the discouraged field/prop <blub>" is better.

Hosch250:
Hmm, seems the old package (which is used in `develop`) isn't available, or something.  I had to update it to the next one.

VSDiagnostics/VSDiagnostics/VSDiagnostics/VSDiagnosticsResources.resx
Hosch250: This message does have a description of what exactly is wrong, except I build it into a string slightly differently than your suggestion. See https://github.com/VSDiagnostics/VSDiagnostics/pull/601/files#diff-42e7dce63789d70b5a21ed78a68cac2aR170

PR #607, opened 2016-08-02
Tweak Flags enum string values

Hosch250:


Vannevelj:
This only adds messages but doesn't change code to use them and they're named badly `1`-suffix. Did you forget to include some changes?

Hosch250:
That's peculiar--all I did was edit the values of the existing strings.

Hosch250:
This is ready to merge.

PR #616, opened 2016-12-30
Field Can Be Made Readonly

### LINKED ISSUE #223: Implement FieldCanBeMadeReadonly
A field that is only written to inside constructors or inline, should be made `readonly`.

Hosch250:
Close #223 

Note: it doesn't seem to be displaying the code fix, for some queer reason.


Hosch250:
This'll need to be redone, I just realized.  Just want to check how AppVeyor does.

Issue #614, opened 2016-12-15
Implement Equals and GetHashCode Code Fix Bug


Hosch250:
I encountered this failure with the following class:

```
namespace CheckersUI
{
    public sealed partial class App
    {
        public App()
        {
            InitializeComponent();
        }
    }
}
```

Vannevelj:
What's the bug?

Hosch250:
I haven't checked yet.  The code fix just has an error.

Hosch250:
It seems that it is picking something up in the `.g.i.cs` file for UWP and WPF projects.  We'll have to ignore these locations in the code fixes as well as when firing analyzers (not when inspecting the code in analyzers, though).

Hosch250:
After further research into this, I think I found the problem's root. We only have the root for one document:

    var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);

This messes up the call when we try to get the type declaration node:

    var node = (TypeDeclarationSyntax) root.FindNode(location.SourceSpan);

So, we just pick up a random node at that location in the document we are in.  This means that rather than checking the node for members, we'll have to check the symbol for members we are interested in.  Hmm, why didn't we do that in the first place?

PR #622, opened 2017-02-12
Field Can Be Readonly

### LINKED ISSUE #223: Implement FieldCanBeMadeReadonly

Hosch250:
Close #223


Hosch250:
Todo: static fields assigned from non-static ctor cannot be readonly.

Hosch250:
Todo: Static readonly field cannot be assigned from non-static ctor.

VSDiagnostics/VSDiagnostics/VSDiagnostics.Test/Tests/General/FieldCanBeReadonlyTests.cs
```diff
@@ -0,0 +1,361 @@
...
+    class MyClass
+    {
+        private readonly int _foo = 0, _bar = 1;
+    }
+}";
+
+            VerifyDiagnostic(original,
+                string.Format(FieldCanBeReadonlyAnalyzer.Rule.MessageFormat.ToString(), "_foo"),
+                string.Format(FieldCanBeReadonlyAnalyzer.Rule.MessageFormat.ToString(), "_bar"));
+            VerifyFix(original, result);
```
JohanLarsson: Why not `VerifyFix(original, original);`? Clearer & shorter IMO.
Or maybe a `VerifyNotFixed(original);`?

Hosch250:
This has an issue with crashing on nodes in generated files in WPF (essentially the same problem as #614).  It can't find the node for some reason or other, but it doesn't matter because we don't need to analyze it.

PR #651, opened 2020-08-28
Fixed: Analyzer name in README file

kmataru:


Vannevelj:
Good catch, thanks 👍 

FWIW, if you're interested in this library I'd recommend checking out https://github.com/Vannevelj/SharpSource which I built as a follow-up. It focuses more on the interesting potential issues

kmataru:
> FWIW, if you're interested in this library I'd recommend checking out https://github.com/Vannevelj/SharpSource which I built as a follow-up. It focuses more on the interesting potential issues

Cool stuff! Thanks for sharing.

